PR #1, opened 2022-06-08
Add package setup

r-ash:
What we want from leapfrog is
* A header only library which can be used by external collaborators (avenir)
* Templated C++ code which can be plugged into [TMB](https://github.com/kaskr/adcomp/wiki)
* An Rcpp wrapper so it can be called from an R package
* Tested, documented and running on CI

I am not expecting this repo to exist forever, just a temporary place for development and then when we have a good setup I will merge into [mrc-ide/leapfrog](https://github.com/mrc-ide/leapfrog)

This PR will
* Setup a placeholder header only library at `inst/include` which can be used by avenir
* Testing for the header only library using catch, setup via `testthat::use_catch()` 
* A placeholder Rcpp wrapper which is tested from R
* Github actions build, code coverage and pkgdown site

Questions
* Can we document the header only library in some way that is easy for avenir to consume? Thinking add it as part of the pkgdown website or something but I've not seen an example like this. Maybe just do comments in the code and they can look at that?
* Is this setup to make the header library available to the Rcpp wrapper sensible? i.e. the `Makevars` and `Makevars.win` files - I am copying this setup from Jeff

For future reference, to set this up I used
* cookie cutter R package (manually removing bits we don't need)
* `usethis::use_rcpp()`
* `testthat::use_catch()`


DESCRIPTION
```diff
@@ -0,0 +1,23 @@
+Package: frogger
+Title: Package Title Here
```
richfitz, line 2: Title, Author, Description all need content

R/zzz.R
```diff
@@ -0,0 +1,3 @@
+.onUnload <- function (libpath) {
+  library.dynam.unload("frogger", libpath)
```
richfitz, line 2: I never use this - what added it? Looks like some packages use it and others don't - I wonder what effect it has...

R/zzz.R
```diff
@@ -0,0 +1,3 @@
+.onUnload <- function (libpath) {
```
richfitz: looks like linting is not set up, I have a feeling that needs to be done by me, I'll do that

inst/include/code.h
```diff
@@ -0,0 +1,8 @@
+#ifndef FROGGER_H
```
richfitz, line 1: If you want to simplify this, you can (epirically at least) use `#pragma once`. See mode for examples (e.g., https://github.com/mrc-ide/mode/blob/main/inst/include/mode/control.hpp)

src/Makevars.win
```diff
@@ -0,0 +1 @@
+PKG_CPPFLAGS = -I../inst/include/
```
richfitz, line 1: this file can be deleted

src/code.cpp
```diff
@@ -0,0 +1,14 @@
+#include <Rcpp.h>
```
richfitz, line 1: This file should not exist if you're making a header-only lib

src/test-example.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+// All test files should include the <testthat.h>
+// header file.
+#include <testthat.h>
+
+// Normally this would be a function from your package's
+// compiled library -- you might instead just include a header
+// file providing the definition, and let R CMD INSTALL
+// handle building and linking.
+int twoPlusTwo() {
```
richfitz, line 18: should you not be testing your adder here?

richfitz: or should this file just be removed?

inst/include/code.h
```diff
@@ -0,0 +1,8 @@
...
+#define FROGGER_H
```
richfitz, line 2: The file extension here should be .hpp as this will be C++ not C

inst/include/code.h
```diff
@@ -0,0 +1,8 @@
...
+
+inline int add(int a, int b) {
```
richfitz, line 4: This `inline` here is important because this is a concrete definition. If you'd done a template function like

```
template <typenameT > 
T add(T a, T b) {
  return a + b;
}
```

you would not need it as templates are implicitly inline. See https://en.cppreference.com/w/cpp/language/inline

This is important in your case because without it, every time you include this header you get another full definition of the function which violates the one-definition rule.

richfitz (commented):
> Can we document the header only library in some way that is easy for avenir to consume? 
> Thinking add it as part of the pkgdown website or something but I've not seen an example like this. Maybe just do comments in the code and they can look at that?

I've done a really terrible job of this with dust's C++ lib, using doxygen. I did get some way through trying to generalise that last winter but it's not done yet. See for example [this code](https://github.com/mrc-ide/dust/blob/7c9e15679af1eec9baf2085da88a47c49076a5ce/inst/include/dust/random/xoshiro_state.hpp#L17-L20) and the [generated docs](https://mrc-ide.github.io/dust/cpp/random-state.html). The bits in between this are not pretty though.

> Is this setup to make the header library available to the Rcpp wrapper sensible? i.e. the Makevars and Makevars.win files - I am copying this setup from Jeff

You do not need the Makevars.win in your case, as it is identical to the Makevars file

> Rcpp vs cpp11 - Jeff mentions some of the syntactic sugar is useful in Rcpp but that he thought there was less good support for documentation? That looks like that might not be the case anymore, do you have any thoughts on Rcpp vs cpp11?

I much prefer cpp11 as it's easier to reason about and is better documented. It also compiles a bit faster. It's possible you will hit issues with it and TMB or Eigen if they are assuming Rcpp.

richfitz (changes requested):
> Can we document the header only library in some way that is easy for avenir to consume? 
> Thinking add it as part of the pkgdown website or something but I've not seen an example like this. Maybe just do comments in the code and they can look at that?

I've done a really terrible job of this with dust's C++ lib, using doxygen. I did get some way through trying to generalise that last winter but it's not done yet. See for example [this code](https://github.com/mrc-ide/dust/blob/7c9e15679af1eec9baf2085da88a47c49076a5ce/inst/include/dust/random/xoshiro_state.hpp#L17-L20) and the [generated docs](https://mrc-ide.github.io/dust/cpp/random-state.html). The bits in between this are not pretty though.

> Is this setup to make the header library available to the Rcpp wrapper sensible? i.e. the Makevars and Makevars.win files - I am copying this setup from Jeff

You do not need the Makevars.win in your case, as it is identical to the Makevars file

> Rcpp vs cpp11 - Jeff mentions some of the syntactic sugar is useful in Rcpp but that he thought there was less good support for documentation? That looks like that might not be the case anymore, do you have any thoughts on Rcpp vs cpp11?

I much prefer cpp11 as it's easier to reason about and is better documented. It also compiles a bit faster. It's possible you will hit issues with it and TMB or Eigen if they are assuming Rcpp.

hillalex (approved):
Can't offer any feedback on the setup as all still unfamiliar to me, but I was able to install and run it :)

R/zzz.R
r-ash: I got this from https://r-pkgs.org/src.html#c-best-practices

R/zzz.R
r-ash: Good catch, will add that here and I've made a maintenance ticket https://mrc-ide.myjetbrains.com/youtrack/issue/RESIDE-299

r-ash:
Thanks for reviews, I think I have addressed most of the comments.

Now you've mentioned that dust documentation it rings a bell, I think I might have reviewed one of those PR's. Might look at adding that in a separate PR.

Re Rcpp vs cpp11 I think I might have recalled the deal breaker, from https://cpp11.r-lib.org/articles/converting.html#incomplete-list-of-rcpp-features-not-included-in-cpp11 "No support for roxygen2 comments" it would be a bit of a pain not to have that I think.

richfitz:
> Re Rcpp vs cpp11 I think I might have recalled the deal breaker, from cpp11.r-lib.org/articles/converting.html#incomplete-list-of-rcpp-features-not-included-in-cpp11 "No support for roxygen2 comments" it would be a bit of a pain not to have that I think.

This is not something I've ever missed. It's very rare that a function written in C++ wants to be directly exported without some degree of wrapping at the R side, which is where the roxygen then tends to live

src/code.cpp
richfitz: oh sorry, I see

richfitz (approved):

PR #2, opened 2022-06-15
mrc-3273: Implement demographic projection

r-ash:
This PR will
* Run demographic projection for just general population

Next steps
* Expand model to run demography for HIV population
* Expand model to include HIV model simulation


tests/testthat/test-frogger.R
```diff
@@ -1,3 +1,41 @@
...
+  demp <- readRDS(test_path("testdata/demographic_projection_object.rds"))
+
+  expect_error(run_base_model(demp, NULL), NA)
+})
+
+test_that("error thrown if trying to run model for more than max years", {
+  demp <- readRDS(test_path("testdata/demographic_projection_object.rds"))
+
+  expect_error(run_base_model(demp, 100L), "No of years > max years of 60")
 })
```
r-ash, line 41: We talked about adding a test here which sets no of women to 0 to check no births. Tried this (and tried setting fertility rate to 0) but this returns NaNs because we use population to find the migration rate. So after running for a few years these NaNs propagate through all the values so not very interesting to test.

inst/include/frogger.hpp
```diff
@@ -0,0 +1,121 @@
+#pragma once
+
+#include <iostream>
+#include "types.hpp"
+
+const int MALE = 0;
```
richfitz, line 6: This is pretty nasty - you'll have a sense of how widepread this is across the code as to if we need to keep these as `#defines`

inst/include/frogger.hpp
```diff
@@ -0,0 +1,121 @@
...
+const int FEMALE = 1;
+
+template <typename real_type>
+State<real_type> model_runner(int time_steps,
+                              const Parameters<real_type>& pars) {
+  State<double> state(pars.age_groups_pop, pars.num_genders);
+  initialise_model_state(pars, state);
+  auto state_next = state;
+  for (int time = 1; time <= time_steps; ++time) {
```
richfitz, line 15: With discrete time models, it's always worth getting pretty concrete about what `time` means here - is this the value at the beginning or at the end of the step?

inst/include/frogger.hpp
```diff
@@ -0,0 +1,121 @@
+#pragma once
+
+#include <iostream>
```
richfitz, line 3: This import can be dropped

src/frogger.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+// [[Rcpp::export]]
+Rcpp::List run_base_model(const Rcpp::List demp, SEXP sim_years) {
+  const int proj_years = get_simulation_years(demp, sim_years);
+  const int num_genders = 2;
+  const int age_groups_pop = 81;
+  const int fertility_first_age_group = 15;
+  const int age_groups_fert = 35;
+
+  TensorMap2<double> base_pop(REAL(demp["basepop"]), age_groups_pop,
```
richfitz, line 32: At some point we will need validation of the sizes here or we'll risk crashes

inst/include/frogger.hpp
r-ash: There are about 7 constants like this that Jeff has in the full model. Should I switch that to a #define?

src/frogger.cpp
r-ash: Thanks, created a ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3301

inst/include/frogger.hpp
r-ash: It's the mid point of the year, I've added a comment for that

r-ash: Oh yeah this is a bit confusing, I tried to comment it to explain in `frogger.cpp` where we set the `max_sim_years`. But probably I should put something in here too.

The initial state is filled in from input data at time index 0 (represents year 1970 in example data). In the loop, at each time step we fill `state_next` from `state` and the input data (e.g. `survival`, `age_sex_fertility_ratio` etc.) at time `time`. So loop runs from 1 as we have used the first years data for initial state. 

I'm then running up to `<= time_steps` because I wanted the `sim_years` arg to `run_base_model` to be the number of times the loop runs. The logic which sets `time_steps` sets max value to be 1 less than the length so we don't run off the end of the input data.

src/RcppExports.cpp
```diff
@@ -10,24 +11,21 @@ Rcpp::Rostream<true>&  Rcpp::Rcout = Rcpp::Rcpp_cout_get();
...
+// run_base_model
+Rcpp::List run_base_model(const Rcpp::List demp, SEXP sim_years);
+RcppExport SEXP _frogger_run_base_model(SEXP dempSEXP, SEXP sim_yearsSEXP) {
 BEGIN_RCPP
     Rcpp::RObject rcpp_result_gen;
     Rcpp::RNGScope rcpp_rngScope_gen;
-    Rcpp::traits::input_parameter< int >::type a(aSEXP);
-    Rcpp::traits::input_parameter< int >::type b(bSEXP);
-    rcpp_result_gen = Rcpp::wrap(adder(a, b));
+    Rcpp::traits::input_parameter< const Rcpp::List >::type demp(dempSEXP);
```
hillalex, line 20: is the whitespace inside angle brackets canonical styling? intellij does this too but I thought it was a bug

richfitz: That was a pretty common style in pre-C++11 style because it helped disambiguate `>>` from `> >`, but is discouraged now. That file is autogenerated file, but in stuff we write no spaces is better

inst/include/frogger.hpp
```diff
@@ -0,0 +1,120 @@
...
+        (state_next.total_population(a, g) +
+         0.5 * state_next.natural_deaths(a, g)) /
+        (state_next.total_population(a, g) + state_next.natural_deaths(a, g));
+    migration_rate(a, g) = survival_netmig *
+                           pars.net_migration(a, g, time_step) /
+                           state_next.total_population(a, g);
+    state_next.total_population(a, g) *= 1.0 + migration_rate(a, g);
+  }
+
+  // fertility
```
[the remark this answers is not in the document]
r-ash: Yeah good point, I've tried to break that up a bit more now. Also added a `working` data object for storing data that we need just within a step for storing some intermediate value between the functions

inst/include/demographic_projection.hpp
```diff
@@ -0,0 +1,113 @@
+#pragma once
+
+#include "types.hpp"
+
+template <typename real_type>
```
richfitz, line 5: later on, as this stuff grows we might use namespaces to try and keep it under control (particularly to make clear which bits are "science" code and which bits are the framework)

richfitz (approved):

PR #3, opened 2022-06-17
mrc-3300: Add HIV population demographic projection

r-ash:
Not much meaningful to test here as the HIV population and deaths is going to stay 0 until we add incidence.

I think I'd like to try and tidy up the different functions here too, not sure if the split into 
* hiv_population_demographic_projection
* hiv_aging_and_mortality
* hiv_and_art_stratified_ageing
* hiv_deaths_and_migration
* hiv_and_art_stratified_deaths_and_migration

is the most useful, but will require some input from Jeff I think to make a sensible split


inst/include/hiv_demographic_projection.hpp
```diff
@@ -0,0 +1,168 @@
...
+        state_next.hiv_strat_adult(hm, ha, g) =
+            (1.0 - hiv_ag_prob(ha, g)) *
+            state_curr.hiv_strat_adult(hm, ha, g);  // age-out
+        state_next.hiv_strat_adult(hm, ha, g) +=
+            hiv_ag_prob(ha - 1, g) *
+            state_curr.hiv_strat_adult(hm, ha - 1, g);  // age-in
+        if (time_step > pars.time_art_start)
+          for (int hu = 0; hu < pars.treatment_stages; hu++) {
+            state_next.art_strat_adult(hu, hm, ha, g) =
+                (1.0 - hiv_ag_prob(ha, g)) *
```
r-ash, line 78: These lines are not run atm. This is because in the example data (which has come from Jeff) `time_art_start` is 60 and max of `time_step` is 60 too so this is never run. I will add a test where I set `time_art_start` lower but just leaving a note here to ask Jeff about this when he is back

src/frogger.cpp
```diff
@@ -21,32 +21,72 @@ int get_simulation_years(const Rcpp::List demp, SEXP r_sim_years) {
...
                                              age_groups_fert, proj_years);
-  TensorMap2<double> births_sex_prop(REAL(demp["births_sex_prop"]), num_genders,
+  TensorMap2<double> births_sex_prop(REAL(data["births_sex_prop"]), num_genders,
                                      proj_years);
 
   Parameters<double> params = {num_genders,
                                age_groups_pop,
                                fertility_first_age_group,
                                age_groups_fert,
+                               age_groups_hiv,
```
r-ash, line 84: This was the line that was triggering the `warning: narrowing conversion of 'age_groups_hiv' from 'long int' to 'int' [-Wnarrowing]` warning. I fixed by changing line 64 from 

```
auto age_groups_hiv = age_groups_hiv_span.size();
```
to
```
int age_groups_hiv = age_groups_hiv_span.size();
```
which makes the cast explicit?

inst/include/demographic_projection.hpp
```diff
@@ -8,18 +8,16 @@ void run_demographic_projection(int time_step,
                                 const State<real_type>& state_curr,
                                 State<real_type>& state_next,
                                 WorkingData<real_type>& working) {
-  run_ageing_and_mortality(time_step, pars, state_curr, state_next, working);
+  run_ageing_and_mortality(time_step, pars, state_curr, state_next);
```
richfitz, line 11: It might be nice to keep the same signature for all steps so that you don't have to think too much about what you're passing around. We could formalise that and make a small struct that keeps all the bits together of course...

This would help later if we decided that we want to loop over some array of pointers to functions

inst/include/hiv_demographic_projection.hpp
```diff
@@ -0,0 +1,168 @@
...
+#include <iostream>
+#include "types.hpp"
+
+template <typename real_type>
+void run_hiv_pop_demographic_projection(int time_step,
+                                        const Parameters<real_type>& pars,
+                                        const State<real_type>& state_curr,
+                                        State<real_type>& state_next,
+                                        WorkingData<real_type>& working) {
+  run_hiv_ageing_and_mortality(time_step, pars, state_curr, state_next);
```
richfitz, line 12: So we have `run_ageing_and_mortality` and `run_hiv_ageing_and_mortality` - this really begs for some discussion of what is going on that is similar or different between these!

src/frogger.cpp
```diff
@@ -21,32 +21,72 @@ int get_simulation_years(const Rcpp::List demp, SEXP r_sim_years) {
...
+                          std::string hiv_age_stratification = "full") {
+  const int proj_years = get_simulation_years(data, sim_years);
   const int num_genders = 2;
   const int age_groups_pop = 81;
   const int fertility_first_age_group = 15;
   const int age_groups_fert = 35;
+  const int disease_stages = 7;
+  const int treatment_stages = 3;
+  const int hiv_adult_first_age_group = 15;
+  const int time_art_start = *INTEGER(projection_parameters["t_ART_start"]) -
```
richfitz, line 60:
```suggestion
  const int time_art_start = Rcpp::as<int>(projection_parameters["t_ART_start"]) - 1;
```

richfitz (changes requested):
I'm a bit confused here - it might be worth either adding some docs as to what is going on with the similar-but-different functions, or changing tack a little.

I've not gone through the functions but it seems like there some parts that are very similar for HIV and non-HIV compartments

```
  for (int g = 0; g < pars.num_genders; ++g) {
    for (int a = 1; a < pars.age_groups_pop; ++a) {
      state_next.hiv_natural_deaths(a, g) =
          state_curr.hiv_population(a - 1, g) *
          (1.0 - pars.survival(a, g, time_step));
      state_next.hiv_population(a, g) = state_curr.hiv_population(a - 1, g);
    }
...
  for (int g = 0; g < pars.num_genders; ++g) {
    // Start at index 1 as we will add infant (age 0) births and deaths later
    for (int a = 1; a < pars.age_groups_pop; ++a) {
      state_next.natural_deaths(a, g) = state_curr.total_population(a - 1, g) *
                                        (1.0 - pars.survival(a, g, time_step));
      state_next.total_population(a, g) =
          state_curr.total_population(a - 1, g) -
          state_next.natural_deaths(a, g);
    }
```

which would be better if they went through the same underlying code. But then there are also splits that look quite different: `run_hiv_deaths_and_migration` vs `run_migration`. Perhaps this can't be sorted out with some refactoring on the science side? But probably worth thinking about and at least coming up with a wishlist and docs if we don't refactor now

inst/include/frogger.hpp
```diff
@@ -1,16 +1,21 @@
 #pragma once
 
 #include "demographic_projection.hpp"
+#include "hiv_demographic_projection.hpp"
```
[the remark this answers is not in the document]
r-ash: The population that it is run on, HIV demographic projection runs on hiv population data, and hiv population stratified by disease stage and art population stratified by disease stage and treatment stage.

The demographic projection `run_ageing_and_mortality` and `run_hiv_ageing_and_mortality` are quite similar you are right. Only difference is the HIV one doesn't remove the natural deaths from the hiv population until later - I think because it uses the hiv population to calculate the stratified populations.

The migration code is quite different because the demographic projection part calculates a migration rate as a proportion of the population size, this is then reused in the HIV demographic projection

r-ash:
I've made the file and function name changes we discussed. Will leave bigger refactoring of functions until after we have reference tests in. At the moment the code is very similar to what Jeff has written and will be hard to manage big changes to the code without the reference tests. I've added a note to the readme with a TODO to cover the refactoring we discussed.

src/frogger.cpp
```diff
@@ -58,20 +99,46 @@ Rcpp::List run_base_model(const Rcpp::List demp, SEXP sim_years) {
   Rcpp::NumericVector r_total_population(age_groups_pop * num_genders);
   Rcpp::NumericVector r_births(1);
   Rcpp::NumericVector r_natural_deaths(age_groups_pop * num_genders);
+  Rcpp::NumericVector r_hiv_population(age_groups_pop * num_genders);
+  Rcpp::NumericVector r_hiv_natural_deaths(age_groups_pop * num_genders);
+  Rcpp::NumericVector r_hiv_strat_adult(disease_stages * age_groups_hiv *
```
r-ash, line 104: Bit confused about how this line is not covered but the others are.. any ideas? Is the coverage for C++ just imperfect?

r-ash: Chatted about this with Rich, looks like codecov with C++ doesn't always produce most accurate reports. Don't need to worry too much about this missed line

src/frogger.cpp
```diff
@@ -58,20 +99,46 @@ Rcpp::List run_base_model(const Rcpp::List demp, SEXP sim_years) {
...
+  r_hiv_strat_adult.attr("dim") =
+      Rcpp::NumericVector::create(disease_stages, age_groups_hiv, num_genders);
+  r_art_strat_adult.attr("dim") = Rcpp::NumericVector::create(
+      treatment_stages, disease_stages, age_groups_hiv, num_genders);
 
   std::copy_n(state.total_population.data(), state.total_population.size(),
               REAL(r_total_population));
   std::copy_n(state.natural_deaths.data(), state.natural_deaths.size(),
               REAL(r_natural_deaths));
+  std::copy_n(state.hiv_population.data(), state.hiv_population.size(),
```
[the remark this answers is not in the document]
r-ash: Added a TODO note for this

inst/include/frogger.hpp
```diff
@@ -1,16 +1,21 @@
 #pragma once
```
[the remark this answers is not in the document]
r-ash: Added a note into the README TODOs

richfitz (approved):

PR #5, opened 2023-05-25
Add leapfrog and internal namespaces

r-ash:
This PR is quite noisy but only has minor changes
* Add a top level namespace leapfrog which exports `run_model` function and types `TensorMapN`, `TensorN`, `Parameters` and `State`. And adds everything else to the `leapfrog.internal` namespace. We could split up this internal namespace.

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -25,7 +25,7 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: pkgdown
+          extra-packages: any::pkgdown, local::.
           needs: website
 
       - name: Deploy package
--- a/inst/include/frogger.hpp
+++ b/inst/include/frogger.hpp
@@ -22,6 +22,7 @@ void initialise_model_state(const Parameters<real_type> &pars,
   state.art_strat_adult.setZero();
   state.births = 0;
 }
+
 }
 
 template<typename real_type>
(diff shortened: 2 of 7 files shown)


.github/workflows/pkgdown.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: pkgdown
+          extra-packages: any::pkgdown, local::.
```
r-ash, line 28: Fixing this whilst I'm as failing to install on main branch. This changed with change from v1 action to v2

inst/include/frogger.hpp
```diff
@@ -3,37 +3,43 @@
 #include "general_demographic_projection.hpp"
 #include "hiv_demographic_projection.hpp"
 
-template <typename real_type>
-State<real_type> run_model(int time_steps, const Parameters<real_type>& pars) {
-  State<real_type> state(pars.age_groups_pop, pars.num_genders,
-                      pars.disease_stages, pars.age_groups_hiv,
-                      pars.treatment_stages);
+namespace leapfrog {
```
richfitz, line 6: prevailing style is generally not to indent namespaces, you should be able to configure that in your editor

richfitz (changes requested):

inst/include/frogger.hpp
r-ash: Ok, I've updated those rules and run the autoformatted. Has updated some other lines like moving the reference from `const Parameters<real_type>& pars` to `const Parameters<real_type> &pars`

richfitz (approved):

PR #9, opened 2023-06-15
mrc 4321: Add standalone C++ runner

r-ash:
This is for an example to Avenir of how to use the model

This adds a few things to help with this
* Data can be serialized from R using `serialize_r_to_tensor` or from C++ using `serialize_tensor` or `serialize_tensor_map`
* Can be deserialized in C++ using `deserialize_tensor` or in R using `deserialize_tensor_to_r`

The script will read in the input data from specified path, write output to specified path. At the moment it is just checking HIV population but I will expand this to check the other outputs.

The PR looks really big but a decent chunk of that is serialized tensor data.


inst/fit_model/fit_model.cpp
```diff
@@ -0,0 +1,223 @@
...
+  const int pAG_INCIDPOP = 35;
+  const int time_art_start = 30;
+
+  // Set working dir whilst reading files
+  std::filesystem::path old_dir = std::filesystem::current_path();
+  std::filesystem::current_path(input_abs);
+  auto exit_handler = [&old_dir]() {
+    std::filesystem::current_path(old_dir);
+  };
+//  std::atexit(exit_handler);
```
r-ash, line 72: This doesn't work here as `atexit` needs a function which returns void, which the lambda doesn't satisfy. Any idea how to get this pattern to work? We don't need it necessarily in this case as we've already converted out output dir into an absolute path so it should always work even if current path has been updated

inst/fit_model/fit_model.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+  leapfrog::Tensor2<double> base_pop_data = serialize::deserialize_tensor<double, 2>(
+      std::string("basepop"));
+  const leapfrog::TensorMap2<double> base_pop = tensor_to_tensor_map<double, 2>(base_pop_data);
+  leapfrog::Tensor3<double> survival_data = serialize::deserialize_tensor<double, 3>(
+      std::string("survival"));
+  const leapfrog::TensorMap3<double> survival = tensor_to_tensor_map<double, 3>(survival_data);
+  leapfrog::Tensor3<double> net_migration_data = serialize::deserialize_tensor<double, 3>(
+      std::string("net_migration"));
+  const leapfrog::TensorMap3<double> net_migration = tensor_to_tensor_map<double, 3>(net_migration_data);
```
r-ash, line 105: I don't think there is a way to do this in 1 is there? `params` object expects a TensorMap but I want this function to own the tensor data (as it has read it from disk). I tried putting this into a single helper which reads the data, creates the Tensor and returns a TensorMap over it but I think that won't work here as the Tensor data is then cleaned up when that helper function returns. So I think we have to do it in 2 bits like this?

I also tried putting all this into a helper function to then just return the `params` object but ran into the same problems with ownership there.

src/frogger.cpp
```diff
@@ -245,3 +246,41 @@ Rcpp::List run_base_model(const Rcpp::List data,
...
+      Rcpp::NumericVector::create(2, 3, 4);
+
+  std::copy_n(test_2d.data(), test_2d.size(), REAL(r_test_2d));
+  std::copy_n(test_3d.data(), test_3d.size(), INTEGER(r_test_3d));
+
+  Rcpp::List ret =
+      Rcpp::List::create(Rcpp::_["test_2d_double"] = r_test_2d,
+                         Rcpp::_["test_3d_int"] = r_test_3d);
+  return ret;
+}
```
r-ash, line 286: These are only used for testing

inst/fit_model/fit_model.cpp
richfitz: apparently (and empirically) you should not use atexit for this. So either move the directory changing into try/catch blocks or split up your main function?

richfitz (changes requested):

inst/fit_model/fit_model.cpp
r-ash: Actually maybe that doesn't matter for now as we write our output using an absolute path which is prepared before this so I think this is ok to just remove this commented out line?

src/frogger.cpp
```diff
@@ -143,7 +143,7 @@ Rcpp::List run_base_model(const Rcpp::List data,
                                                adult_incidence_first_age_group,
                                                pAG_INCIDPOP,
                                                hiv_steps,
-                                               (1.0 / hiv_steps),
+                                               dt,
```
r-ash, line 146: Opportunistic tidy

inst/fit_model/fit_model.cpp
```diff
@@ -0,0 +1,223 @@
...
+  const leapfrog::TensorMap1<int> age_groups_hiv_span = tensor_to_tensor_map<int, 1>(age_groups_hiv_span_data);
+  int age_groups_hiv = static_cast<int>(age_groups_hiv_span.size());
+  int age_groups_hiv_15plus = age_groups_hiv;
+  const int scale_cd4_mortality = 1;
+  int hIDX_15PLUS = 0;
+  const double art_alloc_mxweight = 0.2;
+
+  leapfrog::Tensor1<int> v = serialize::deserialize_tensor<int, 1>(std::string("artcd4elig_idx"));
+  for (int i = 0; i <= sim_years; ++i) {
+    // 0-based indexing in C++ vs 1-based indexing in R
```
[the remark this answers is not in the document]
richfitz: cool, so long as it ends up very clear whose responsibility this is and its very consistent

richfitz (approved):

PR #11, opened 2023-06-20
README.md naming conventions -- WIP

jeffeaton:
Next:

* Rob to propose variable name conventions + new names
* Maggie help split internal components  (split based on whether its in the demo bit or in the disease progression model)

PR #12, opened 2023-06-22
mrc-4339: Add a state saver object to configure returned years

r-ash:
This PR will enable callers to save output from more than 1 timestep. The interface I've gone for here is users setup a vector of indices they want to output e.g. `0:60` for all years. You initialise the `StateSaver` with this setup then that handles allocating the memory and copying output after every time step into output. I wonder if that controlling of whether to save a particular year should move into the controlling process, I think it might be a bit more obvious what it is doing if we move it there.


src/frogger.cpp
```diff
@@ -54,13 +54,19 @@ leapfrog::TensorMap1<int> get_age_groups_hiv_span(const Rcpp::List projection_pa
   return age_groups_hiv_span;
 }
 
+std::vector<int> parse_output_steps(Rcpp::NumericVector output_steps, int proj_years) {
```
richfitz, line 57: what is being parsed here? why is `proj_years` an input that is not used? should you also verify that these are (strictly) increasing?

richfitz (changes requested):

inst/include/state_saver.hpp
```diff
@@ -0,0 +1,118 @@
...
+    Tensor4<real_type> hiv_strat_adult;
+    Tensor5<real_type> art_strat_adult;
+    Tensor1<real_type> births;
+    Tensor4<real_type> aids_deaths_no_art;
+    Tensor3<real_type> infections;
+    Tensor5<real_type> aids_deaths_art;
+    Tensor4<real_type> art_initiation;
+    Tensor3<real_type> hiv_deaths;
+
+    OutputState(int age_groups_pop,
```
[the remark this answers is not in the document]
r-ash: Yes they are, so these are all the dimensions of our state-space. We've started thinking about splitting this out in https://github.com/mrc-ide/frogger/pull/11/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R51 as part of that PR I will update the names and also create some sub-struct for the parameters. If it looks ok I would wait until that PR is pinned down and use whatever struct comes from that. I'll add a TODO note here into the README

src/frogger.cpp
r-ash: We don't need proj_years - good catch. I think that was just some old code left around as I originally had this interface take a time step and was then building `output_steps` from the time step and `proj_years`. By parsing here I just really mean converting from R type to C++ type. I've tried updating that name, this is now `transform_output_steps` and I've updated the other functions similar to this one which are also converting from R type to C++ type, possibly doing some validation on the way.

I can check they are increasing but I think this will work atm even if they aren't. If someone passed in `seq(60, 0, -1)` it would return with time backwards if someone really wanted. I can explicitly disallow if you think it makes interface tidier

richfitz (approved):

PR #13, opened 2023-06-23
Add scripts to profile frogger code

r-ash:
This adds 3 scripts which can be used for profiling, 1 using valgrind, 1 with google gperftools and 1 using perf. I'm not really sure what is better here! But this should at least give us some options going forward. See the README for details of how to install them and run the script. I've only testing this on linux and its possible we might need a completely different approach for profiling on another OS.


inst/fit_model/fit_model.cpp
```diff
@@ -222,37 +222,48 @@ int main(int argc, char *argv[]) {
...
-  for (int step = 1; step <= sim_years; ++step) {
-    state_next.reset();
-    leapfrog::run_general_pop_demographic_projection(step, params, state_current, state_next, intermediate);
-    leapfrog::run_hiv_pop_demographic_projection(step, params, state_current, state_next, intermediate);
-    leapfrog::run_hiv_model_simulation(step, params, state_current, state_next, intermediate);
-    state_output.save_state(state_next, step);
-    std::swap(state_current, state_next);
-    intermediate.reset();
+
+  const char *is_profiling = std::getenv("PROFILING");
```
richfitz, line 239: why not pass this (n) as an arg with default 0?

r-ash: Yes I don't know! No good reason, let me update that

richfitz (approved):

PR #14, opened 2023-06-27
mrc-4346: As a user I can run a coarse stratified model fit from R

r-ash:
This builds on #12 and should be reviewed after that

This PR will
* Add ability to run a coarse age stratified model fit from R
* Smaller helper function to parse data which will error if it is not long enough and so could lead to accessing unallocated memory


src/frogger.cpp
```diff
@@ -31,40 +31,86 @@ int get_hiv_steps_per_year(SEXP r_hiv_steps_per_year) {
...
   return Rcpp::as<std::vector<int>>(output_steps);
 }
 
+void validate_stratification(std::string stratification) {
+  if (!(stratification == "full" || stratification == "coarse")) {
+    Rcpp::stop(
+        "Invalid HIV age stratification must be 'full' or 'coarse' got '%s'.",
+        stratification);
+  }
+}
```
r-ash, line 104: I also toyed around with validating and returning an enum but then where I want to use this I want to append `_full` or `_coarse` on the end of the string so felt a bit weird to convert from the string to an enum and back again to a string again so I removed that

src/frogger.cpp
```diff
@@ -82,59 +125,68 @@ Rcpp::List run_base_model(const Rcpp::List data,
...
+
   int age_groups_hiv_15plus = age_groups_hiv;
   const int scale_cd4_mortality =
       Rcpp::as<int>(projection_parameters["scale_cd4_mort"]);
   int hIDX_15PLUS = 0;
   const double art_alloc_mxweight = Rcpp::as<double>(projection_parameters["art_alloc_mxweight"]);
 
-  const leapfrog::TensorMap2<double> base_pop(REAL(data["basepop"]), age_groups_pop,
-                                              num_genders);
+  std::function<std::string(const std::string&)> build_prop_name = [hiv_age_stratification](const std::string& property) {
```
richfitz, line 136: why is this a lambda? more generally, why do we need to do string manipulation here? I would generally be suggesting moving all of this sort of thing into the R layer, as this is likely to only get more painful

src/frogger.cpp
```diff
@@ -31,40 +31,83 @@ int get_hiv_steps_per_year(SEXP r_hiv_steps_per_year) {
...
+
+template<typename... Args>
+auto parse_data_int(const Rcpp::List data, const std::string& key, Args... dims) {
+  constexpr std::size_t rank = sizeof...(dims);
+  Eigen::array<int, rank> dimensions{ static_cast<int>(dims)... };
+
+  int product = 1;
+  for (size_t i = 0; i < rank; ++i) {
+    product *= dimensions[i];
   }
```
richfitz, line 53:
```suggestion
  const int product = std::accumulate(dimensions.begin(), dimensions.end(), 1, std::multiplies<size_t>());
```

See https://en.cppreference.com/w/cpp/algorithm/accumulate

There are _some_ nice bits in C++ :)

richfitz (changes requested):
Some scattered comments, but in general probably need to be looped back in on this I am afraid.

src/frogger.cpp
```diff
@@ -31,40 +31,86 @@ int get_hiv_steps_per_year(SEXP r_hiv_steps_per_year) {
...
+               key,
+               product,
+               LENGTH(array_data));
+  }
+
+  return Eigen::TensorMap<Eigen::Tensor<int, rank>>(INTEGER(array_data), static_cast<int>(dims)...);
+}
+
+template<typename... Args>
+auto parse_data_double(const Rcpp::List data, const std::string& key, Args... dims) {
```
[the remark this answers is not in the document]
r-ash: Just what I wanted! Thanks!

src/frogger.cpp
```diff
@@ -82,59 +125,68 @@ Rcpp::List run_base_model(const Rcpp::List data,
...
+      projection_parameters, build_prop_name("cd4_initdist"), disease_stages, age_groups_hiv, num_genders);
+  const leapfrog::TensorMap1<int> hiv_age_groups_span = parse_data_int(
+      projection_parameters, build_prop_name("hAG_SPAN"), age_groups_hiv);
+  const leapfrog::TensorMap4<double> art_mortality = parse_data_double(
+      projection_parameters, build_prop_name("art_mort"), treatment_stages, disease_stages, age_groups_hiv,
+      num_genders);
+  const leapfrog::TensorMap2<double> artmx_timerr = parse_data_double(projection_parameters, "artmx_timerr",
+                                                                      treatment_stages, proj_years);
   leapfrog::Tensor1<double> h_art_stage_dur(treatment_stages - 1);
   for (int i = 0; i < treatment_stages - 1; ++i) {
```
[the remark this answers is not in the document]
r-ash: ah yeah there is a `setConstant` function!

src/frogger.cpp
r-ash: Writing this out in some detail for my understanding too.

In the full model there are 66 ages in the HIV model. Using Jeff's variable names, these start at age `pIDX_HIVADULT` (= `15`) for `hAG` (= `66`) ages. Meaning the HIV part of the model runs from age 15 up to the 80+ open ended age group.

There are 2 ways to run the model, with coarse age stratified or full age stratified HIV population. If fully stratified input data is given for all the individual 66 HIV age groups e.g. age = 15, age = 16, age = 17, .... If we run coarse stratified input data is supplied for a range of age groups. With the current input data this is age ranges 15-16, 17-19, 20-24, 25-29, ... . 

The coarse age stratified model is used for ? and the full model for ? (will double check with Jeff)

At the moment here and in leapfrog all the input data is passed in, so we pass in e.g. `art_mort_full` and `art_mort_coarse` (both the coarse and full stratified input data) and an arg `hiv_age_stratification` which controls whether we use `art_mort_full` or `art_mort_coarse` etc. The names of all these inputs follow a common pattern `prop_full` or `prop_coarse` hence the string manipulation to pick the correct data based on the value of `hiv_age_stratification`.

I think you're right we can move this into R as it would keep the C++ wrapper a bit cleaner which is always nice! Will just double check with Jeff when we want to use the different types of fit

richfitz (approved):

PR #19, opened 2023-07-05
mrc-4362: Enumerate the possible state spaces and statically allocate vectors

r-ash:
This PR will
* Add a struct for storing the possible state spaces and template the model on this
* Use TensorFixedSizes where possible (this gives a small speedup)

This is quite a noisy change as the new state space struct used throughout the code


inst/include/frogger.hpp
```diff
@@ -32,32 +34,27 @@ void initialise_model_state(const Parameters<real_type> &pars,
...
   // Save initial state
   state_output.save_state(state, 0);
 
   // Each time step is mid-point of the year
   for (int step = 1; step <= time_steps; ++step) {
     state_next.reset();
-    run_general_pop_demographic_projection(step, pars, state, state_next, intermediate);
-    run_hiv_pop_demographic_projection(step, pars, state, state_next, intermediate);
-    run_hiv_model_simulation(step, pars, state, state_next, intermediate);
+    run_general_pop_demographic_projection<real_type, S>(step, pars, state, state_next, intermediate);
```
richfitz, line 55: You probably don't want to hear this, but I think that if you reorder the template parameters you may be able to drop the `real_type` here and just template explicitly on `S`. I forget if this works with mixed type/literal templates though

richfitz (approved):

inst/include/frogger.hpp
r-ash: Sounds good, I've tidied that up

PR #21, opened 2023-07-18
Paediatric ageing

mwalte10:
This PR implements paediatric ageing by introducing a function within the child model script.


mwalte10:
Rob if you could also help me figure out why my checks are failing that would be great!

r-ash:
> Rob if you could also help me figure out why my checks are failing that would be great!

Failing because of the standalone runner. This is a test which runs the script here https://github.com/mrc-ide/frogger/blob/main/inst/fit_model/fit_model.cpp which runs the model without using R. So is the example for avenir of how to drive this from C++ alone. Failing because it expects the child data to exist at the moment. I need to fix this in mrc-4390 branch that this PR builds off.

Note I've also updated the base branch here to mrc-4390. This is just to make the diff easier to see for review but let's switch it back to main before merging.

re code factor issues see https://www.codefactor.io/repository/github/mrc-ide/frogger/pull/21 these run [lintr](https://lintr.r-lib.org/) for R and [google code style](https://google.github.io/styleguide/cppguide.html) for C++ to ensure code conforms to a style for ease of reading. Re the complex method warning though that's a bit more fiddly.

inst/include/child_model_simulation.hpp
```diff
@@ -6,6 +6,65 @@ namespace leapfrog {
...
+        }
+        for(int dur = 0; dur < ss.treatment_stages; ++dur){
+          state_next.hc_art_pop(dur, hm_alt, 5, g) += state_curr.hc_art_pop(dur, hm, 4, g) * demog.survival(5, g, time_step) * children.hc_cd4_transition(hm_alt, hm);
+        }
+      }
+    }
+  }
+  
+  for(int g = 0; g < ss.num_genders; ++g){
+    //thought it was ss.fertility_first_age_group but maybe not? ASK ROB
```
r-ash, line 49: This is in the options at the moment `pars.options.fertility_first_age_group`

tests/testthat/test-child-model.R
```diff
@@ -12,7 +12,7 @@ test_that("child model can be run for all years", {
   expect_true(all(out$hiv_population[1:5, , which(1970:2030 == 1980)] == 10))
   expect_true(all(out$hc_hiv_pop[1, 1, 1:5, , which(1970:2030 == 1980)] == 10))
 
-  ## Only seeding nosocomial infections in year 1980, infections every
-  ## other year should be 0
-  expect_true(all(out$infections[1:15, , which(1970:2030 != 1980)] == 0))
+  ## HIV population should be larger than zero in age 6 in 1981 because ageing is allowed
+  expect_true(all(out$hc_hiv_pop[1, 1, 6, , which(1970:2030 == 1981)] > 0))
```
r-ash, line 16: Nice!

inst/include/child_model_simulation.hpp
```diff
@@ -6,6 +6,65 @@ namespace leapfrog {
...
+    for(int ha = 6; ha < 15; ++ha) {
```
r-ash, line 50: Why does this start at 6 and end at 15? Can we relate this to hC1_disease_stages or hC2_disease_stages at all?

r-ash: Ah actually I think I can see what is happening, we have slightly different behaviour for 1-4, 5 and 6-14

r-ash: It looks like this is doing the same thing for 1-4 and 6 - 14 but slightly differently for 5. Could add a function for the duplicated code on 1-4 and 6-14 but up to you.

r-ash (commented):
This all looks great, some tiny comments but mostly styling related!

inst/include/child_model_simulation.hpp
mwalte10: I think leaving it outside of a function allows you to see more why the aging process is done like this? So I'm more inclined to keep it separate.

inst/include/child_model_simulation.hpp
mwalte10: To clarify pars.options.fertility_first_age_group is equal to 15?

mwalte10:
Merging as a part of #26

PR #32, opened 2023-08-15
Split intermediate data into base and children data

r-ash:
Fair bit of churn, interesting bit is in types.hpp


richfitz (approved):
LGTM, churn is at least easy to follow

PR #34, opened 2023-09-05
Updates following leapfrog meeting 05/09/23

r-ash:
This PR will
* Rename `fit_model` to `simulate_model` in functions
* Rename `fit_model` directory to `standalone_model` I hope this is a bit less confusing naming now
* Move `scale_cd4_mort` and `art_alloc_mxweight` out of options struct and into natural history and art structs

@mwalte10 I hope this isn't too disruptive for your changes - we can look at this after your changes are in if they potentially are


tests/testthat/test-standalone.R
```diff
@@ -18,9 +18,9 @@ test_that("We can compile the standalone program", {
   expect_equal(code, 0)
 
   output <- tempfile()
-  input <- frogger_file("fit_model/data")
+  input <- frogger_file("standalone_model/data")
```
mwalte10, line 21: Trying to understand the nuance between standalone and simulate, is standalone the name of the input object and simulate is what you will call the function?

mwalte10 (approved):
Looks good, just one question in the comments. Once you merge I will pull, merge into my paed branch, and then push

tests/testthat/test-standalone.R
r-ash: Yes opportunistic renaming here. Standalone is just used as a directory name, so only a name used for organising, not a function or an object in the code. This just refers to the directory where the I've written the example of running the model entirely in C++, with no R data or wrapper involved. So it is standalone in that it is standalone C++.

PR #41, opened 2023-10-07
Mkw bf trans

mwalte10:
Incorporates all types of vertical transmission. Some things in particular it would be helpful to get feedback on: 

- Commenting in the functions, too much/little? Trying to document the things that I've found can't be found in the manual or spectrum software
- is the run_child_hiv_infections function in child_model_simulation.hpp clear? trying to functionalize the different parts of vertical transmission.
- Related to the above, what are thoughts on separating some of these functions into different files? Thinking one that contains all the functions for infections, one for natural history, and one for demographic processes. 
- I've modified run_hiv_and_art_stratified_ageing in the demographic file extensively, feel free to suggest any changes to clarify these things!


inst/include/child_model_simulation.hpp
```diff
@@ -66,6 +66,360 @@ void run_child_ageing(int time_step,
...
+  constexpr auto hc_ss = StateSpace<ModelVariant>().children;
+
+
+  //bf_duration is % of women no longer breast feeding
+  for (int bf = 0; bf < hc_ss.hBF; ++bf) {
+    //CHECK: intermediate.children.incidence_rate_wlhiv is coming from perinatal transmission, are there
+    // any checks I should do to ensure that this is reading in properly from the previous function?
+    intermediate.children.bf_at_risk += intermediate.children.incidence_rate_wlhiv / 12 * 2 * (1 - cpars.breastfeeding_duration_no_art(bf, time_step));
+  }
+  // intermediate.children.bf_incident_hiv_transmission_rate = bf_at_risk * 0.269;
```
r-ash, line 335: Now redundant or something which should be on a switch?

inst/include/child_model_simulation.hpp
```diff
@@ -91,10 +446,71 @@ void run_child_hiv_infections(int time_step,
...
       }
-    }
-  }
+    } //end a
+  } // end NS
+
+  internal::run_calculate_perinatal_transmission_rate(time_step, pars, state_curr, state_next, intermediate);
+  //Perinatal transmission
+  for (int s = 0; s < ss.NS; ++s) {
+  for (int hd = 0; hd < hc_ss.hc1DS; ++hd) {
```
r-ash, line 457:
```suggestion
    for (int hd = 0; hd < hc_ss.hc1DS; ++hd) {
```

inst/include/child_model_simulation.hpp
```diff
@@ -66,6 +66,360 @@ void run_child_ageing(int time_step,
...
+  intermediate.children.num_wlhiv = 0.0;
+  intermediate.children.prop_wlhiv_lt200 = 0.0;
+  intermediate.children.prop_wlhiv_200to350 = 0.0;
+  intermediate.children.prop_wlhiv_gte350 = 0.0;
+  intermediate.children.prop_wlhiv_lt350 = 0.0;
+
+  for (int a = 0; a < 35; ++a) {
+    intermediate.children.num_wlhiv_lt200 += state_next.base.h_hiv_adult(4,a,1) + state_next.base.h_hiv_adult(5,a,1) + state_next.base.h_hiv_adult(6,a,1) ;
+    intermediate.children.num_wlhiv_200to350 += state_next.base.h_hiv_adult(3,a,1) + state_next.base.h_hiv_adult(2,a,1) ;
+    intermediate.children.num_wlhiv_gte350 += state_next.base.h_hiv_adult(0,a,1) + state_next.base.h_hiv_adult(1,a,1) ;
```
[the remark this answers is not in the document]
r-ash: Oh thinking about this more that is just picking the bits from the compartments which match to those cd4 counts?

inst/include/child_model_simulation.hpp
```diff
@@ -66,6 +66,360 @@ void run_child_ageing(int time_step,
...
+                                    const State<ModelVariant, real_type> &state_curr,
+                                    State<ModelVariant, real_type> &state_next,
+                                    IntermediateData<ModelVariant, real_type> &intermediate) {
+  const auto cpars = pars.children.children;
+  static_assert(ModelVariant::run_child_model,
+                "run_hiv_child_infections can only be called for model variants where run_child_model is true");
+
+  //Option A and B were only authorized for women with greater than 350 CD4, so if the percentage of women
+  //on option A/B > the proportion of women in this cd4 category, we assume that some must have a cd4 less than 350
+  //option AB will be less effective for these women so we adjust for that
```
r-ash, line 141: This kind of explanatory comment about why we are doing certain things is great

inst/include/child_model_simulation.hpp
```diff
@@ -91,10 +446,71 @@ void run_child_hiv_infections(int time_step,
...
+    } //end a
+  } // end NS
+
+  internal::run_calculate_perinatal_transmission_rate(time_step, pars, state_curr, state_next, intermediate);
+  //Perinatal transmission
+  for (int s = 0; s < ss.NS; ++s) {
+  for (int hd = 0; hd < hc_ss.hc1DS; ++hd) {
+       state_next.children.hc1_hiv_pop(hd, 0, 0, s) +=  state_next.children.hiv_births * intermediate.children.perinatal_transmission_rate * demog.births_sex_prop(s,time_step) * cpars.hc1_cd4_dist(hd);
+      }// end hc1DS
+      state_next.base.p_hiv_pop(0, s) +=  state_next.children.hiv_births * intermediate.children.perinatal_transmission_rate * demog.births_sex_prop(s,time_step);
```
r-ash, line 460: I think I'd be tempted to put
```
state_next.children.hiv_births * intermediate.children.perinatal_transmission_rate * demog.births_sex_prop(s,time_step)
```
into some temporary variable as you use it 3 times here

inst/include/child_model_simulation.hpp
```diff
@@ -91,10 +446,71 @@ void run_child_hiv_infections(int time_step,
...
+      state_next.base.p_infections(0, s) += state_next.children.hiv_births * intermediate.children.perinatal_transmission_rate * demog.births_sex_prop(s,time_step);
+
+  }// end NS
+
+
+   //Breastfeeding transmission
```
r-ash, line 466: Question about definitions, is breastfeeding not part of perinatal transmission? Or does perinatal just cover during pregnancy?

inst/include/child_model_simulation.hpp
```diff
@@ -91,10 +446,71 @@ void run_child_hiv_infections(int time_step,
...
+
+   //0-6
+   internal::adjust_optAB_bf_transmission_rate(time_step, pars, state_curr, state_next, intermediate);
+    internal::run_calculate_transmission_from_incidence_during_breastfeeding(time_step, pars, state_curr, state_next, intermediate);
+    internal::run_bf_transmission_rate(time_step, pars, intermediate, 0, 3, 0);
```
r-ash, line 471:
```suggestion
   internal::adjust_optAB_bf_transmission_rate(time_step, pars, state_curr, state_next, intermediate);
   internal::run_calculate_transmission_from_incidence_during_breastfeeding(time_step, pars, state_curr, state_next, intermediate);
   internal::run_bf_transmission_rate(time_step, pars, intermediate, 0, 3, 0);
```

inst/include/child_model_simulation.hpp
```diff
@@ -91,10 +446,71 @@ void run_child_hiv_infections(int time_step,
...
+   for (int s = 0; s < ss.NS; ++s) {
+     for (int hd = 0; hd < hc_ss.hc1DS; ++hd) {
+       state_next.children.hc1_hiv_pop(hd, 1, 0, s) +=  state_next.children.hiv_births *  demog.births_sex_prop(s,time_step) * cpars.hc1_cd4_dist(hd) * (intermediate.children.bf_incident_hiv_transmission_rate + intermediate.children.bf_transmission_rate(0));
+     }// end hc1DS
+     state_next.base.p_hiv_pop(0, s) +=  state_next.children.hiv_births  * demog.births_sex_prop(s,time_step) * (intermediate.children.bf_incident_hiv_transmission_rate + intermediate.children.bf_transmission_rate(0));
+     state_next.base.p_infections(0, s) += state_next.children.hiv_births  * demog.births_sex_prop(s,time_step) * (intermediate.children.bf_incident_hiv_transmission_rate + intermediate.children.bf_transmission_rate(0));
+   }// end NS
+
+   //6-12
```
r-ash, line 480: I think this reads quite nicely. You could split this function up further into separate 0-6, 6-12, 12+ but I'm not sure if that would really make it easier to read. 

The deeply nested names in this context can obfuscate the actual content though a bit, I might have a look at namespacing those at the top of the function and seeing how much that tidies it up. But I can have a look at that in a separate PR later.

r-ash (changes requested):
This looks really solid, couple of small things I noted but not big changes suggested from me. Couple of things I was thinking about reading through this

* The nested structs do obfuscate the code a fair bit, I'll have a look at collapsing those using some local aliases in another PR at some point
* We should work out how to write documentation for our functions which can be used by consumers of the library

> * Commenting in the functions, too much/little? Trying to document the things that I've found can't be found in the manual or spectrum software

This is really tricky thing to get right and I think everyone ends up settling at a different point. The general guidance from software I try and follow is to try and use comments sparingly. This is because they tend to get out of sync with the code in a codebase. People often update the code and don't remember to edit the comments themselves. You should use the to explain "why" of the code and not what the code is doing. Tell the code reader why you have to do it in this certain way not what you are doing as the code itself should be easy enough to read for them to understand what it is doing and the code itself can't rot like the comments can.

There is an idea of writing "self documenting" code in software development which focuses on using functions and variable names to make the code as readable as possible. There are many articles online about it and it is a principle I try and stick to.

That said a lot of this approach comes out of software development for software which I think is fundamentally different from the kind of code we are writing here. The code here has a lot more dense business logic and less about engineering. So it can be hard to think about where to apply rules of thumb from software world here. 

All that said - basically it is difficult and there is no perfect approach for this kind of code where the meaning in each line if very dense. I think you've struck a fairly good balance from my reading. Using functions where they help separate the logic and aid reuse whilst not going too far to make it a load of confusing spaghetti. Really a good test of the readability might be someone else from the research group working through this code and seeing if they understand what the model is doing.

> is the run_child_hiv_infections function in child_model_simulation.hpp clear? trying to functionalize the different parts of vertical transmission. Related to the above, what are thoughts on separating some of these functions into different files? Thinking one that contains all the functions for infections, one for natural history, and one for demographic processes.

The function looks good to me, I think it has a good balance of functions without splitting the code up too much. I think splitting up some files is a great idea. Putting these into a subdirectory like you mentioned in a call I think is a great shout too. We could have subdirectories for the different model variants. So one for the child model and then you can have a file for each of parts of the model (infections, natural history, demographic) like you suggest.

scripts/maggie_scratch/leapfrog_frogger_testing.R
```diff
@@ -0,0 +1,144 @@
+rm(list = ls())
+library(tidyr)
+library(tidyverse)
+library(dplyr)
+setwd('C:/Users/mwalters/leapfrog/')
```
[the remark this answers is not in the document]
mwalte10: This was mostly just for me so that I could see how I was testing it against spectrum without pulling in the spectrum file. It doesn't need to be pushed

PR #42, opened 2023-10-20
Add state types into code generation

r-ash:
This PR will move the state types into the code generation

In #38 we added code generation for the parameter types i.e. the input types. This adds code generation for the State object. So after this PR if you want to include a new output you have to
1. Add it into model_output.csv with a new column to say which `model_variant` this should be included in. where `ModelVariant` means just use the template name.. This looks a bit redundant with the `output_when` column. I will have another look at this
2. Run the generate script
3. Add the new type into the StateSaver (this is the next step for code generation)
4. Should be ready to use

If you want to add a new variant this will require some additional work to update the templates, but I think that is fine.


M-Kusumgar (commented):
Mainly looks good to me, i think theres a couple places you can use `paste_lines` but I had a quick question actually. The dimensions that you specify in the `model_output.csv` seem unnecessary, you ditch them in the code so why even specify them in the csv? For example
```csv
hc1_art_pop,hc1_art_pop,REAL,ChildModel,ModelVariant::run_child_model,5,base.hTS,children.hc1DS,children.hc1AG,base.NS,output_years
```
this line has `base.hTS` and `children.hc1DS` but they end up as `hTS<ChildModel>` and `hc1DS<ChildModel>` anyway

Edit: I see the difference between  the two in some of the other code generation so only the `paste_lines` comment stands

M-Kusumgar (approved):

PR #43, opened 2023-12-13
Implement end-year net migration

jeffeaton:
Merge #83 before merging this

WIP: _implements change to demographic model; appreciate feedback on pattern_

* ~Mid-year net-migration code commented -- needs to be removed or put under option flag (as in EPP-ASM)~
  * _UPDATE 29 Dec 2023:_ Mid-year net migration implemented with feature flag `projection_period`, same as leapfrog and EPP-ASM code.
* ~Implementation not tested for HIV population~
   * _UPDATE 29 Dec 2023:_ Implementation for net migration among HIV population verified against Spectrum v6.28 test file with migration and no ART (aligned with leapfrog implementation)


jeffeaton:
I am copying this discussion from Teams to hear regarding guidance on implementing migration option in frogger. On 13 December 2023, @r-ash gave on Teams:


> Re frogger impl, when will people switch between these different migration types? Atm in frogger have 3 model variants
> 1. Fine grained ages with child model switched off
> 2. Coarse grained ages with child model off
> 3. Fine grained ages with child model on
> 
> If this is a switch required for all model variants, then best to put it into the model options and switch on it like you are doing in leapfrog. 
> In which case you'll need to
> * Add it to the Options type https://github.com/mrc-ide/frogger/blob/main/inst/include/types.hpp#L79
> * Ensure it is set here https://github.com/mrc-ide/frogger/blob/main/src/frogger.cpp#L48, can use code like > * https://github.com/mrc-ide/leapfrog/pull/52/files#diff-75e479251a20b3b9a839838ebff9f0a524b8cfc095b171b02a9d4792938e1ffbR51-R58 to set that option


@jeffeaton responded:

> Thanks both. Ashton, Robert -- I went back and forth last night about whether the 'end-year migration' should be a ModelVariant or just switches. I provisionally decided that switches were better (not Variant) because it can be set at run-time, vs. everything specified by ModelVariant are compile-time options.  Does that sound right to you?

@r-ash responded:

> Yep that sounds right to me, I think setting at run time will be simpler if we want this switch to be available for every model variant.

r-ash:
Maggie says that we still need to implement the end-year net migration for on ART adult population

jeffeaton:
> Maggie says that we still need to implement the end-year net migration for on ART adult population

I'm not 100% sure what this refers to (suspect we've discussed it before and I have forgotten).  But end-year net migration for the ART adult population I believe is done in this line: https://github.com/mrc-ide/frogger/blob/22f3287806bd578a8e968f2d9f1e186a3c1fb508/inst/include/hiv_demographic_projection.hpp#L240-L242

If there is something else that you recall on here, please let me know

jeffeaton:
Hi @mwalte10, @r-ash,

I have just added a couple of remaining pending things on this:
* Merged the current main branch
* Addressed the difference in Spectrum and EPP-ASM / Leapfrog ART initiation in #83 -- which is merged into this PR
* Accounted for calendar year projection in the ART initiation

I believe that this branch is now up to date with mrc-ide/leapfrog and I think should be ready for testing and merging

This PR now builds on #83, which is smaller; so I suggest reviewing and merging #83 first, and then coming onto this.

mwalte10:
> > Maggie says that we still need to implement the end-year net migration for on ART adult population
> 
> I'm not 100% sure what this refers to (suspect we've discussed it before and I have forgotten). But end-year net migration for the ART adult population I believe is done in this line:
> 
> https://github.com/mrc-ide/frogger/blob/22f3287806bd578a8e968f2d9f1e186a3c1fb508/inst/include/hiv_demographic_projection.hpp#L240-L242
> 
> If there is something else that you recall on here, please let me know

From reviewing this, I think this was probably just me flagging that the ART population needed to be fixed, but that is all sorted now with the other PR.

inst/include/general_demographic_projection.hpp
```diff
@@ -102,12 +125,14 @@ void run_fertility_and_infant_migration(int time_step,
...
+    if (pars.base.options.proj_period_int == internal::PROJPERIOD_MIDYEAR) {
     // Assume 2/3 survival_probability rate since mortality in first six months higher
     // than second 6 months (Spectrum manual, section 6.2.7.4)
-    real_type migration_rate_a0 = demog.net_migration(0, g, time_step) *
-                                  (1.0 + 2.0 * demog.survival_probability(0, g, time_step)) /
-                                  3.0 / state_next.base.p_total_pop(0, g);
-    state_next.base.p_total_pop(0, g) *= 1.0 + migration_rate_a0;
+      real_type migration_rate_a0 = demog.net_migration(0, g, time_step) *
+                                    (1.0 + 2.0 * demog.survival_probability(0, g, time_step)) /
+                                    3.0 / state_next.base.p_total_pop(0, g);
```
mwalte10, line 133: I think that more explanation / a clearer way of presenting this may be useful. I looked at the spectrum manual and I think that changing the note to read as  "// Assume 2/3 survival_probability rate since mortality in first six months of life for children < 1 is higher  // than second 6 months of life (Spectrum manual, section 6.2.7.4)"

For 131-133: 
I'm reading this as:
mig * (1/3 + 2/3 * surv) / pop

Is survival_probability actually a mortality probability? What does the 1/3 represent? Why doesn't this have to be averaged with the year before if it is a mid year calculation?

inst/include/intermediate_data.hpp
```diff
@@ -89,6 +89,9 @@ const int MALE = 0;
 const int FEMALE = 1;
 const int ART0MOS = 0;
 
+const int PROJPERIOD_MIDYEAR = 0;   // mid-year projection period 
+const int PROJPERIOD_CALENDAR = 1;  // calendar-year projection (Spectrum 6.2 update; December 2022)
```
mwalte10, line 93: should the calendar year be zero? I conceptualize 0 as the default, and seems that calendar year will be default going forward

inst/include/model_simulation.hpp
```diff
@@ -239,87 +239,129 @@ void run_h_art_initiation(int hiv_step,
...
           real_type prop_elig = 1.0;
-          intermediate.base.artelig_hahm(hm, ha - hIDX_15PLUS) =
-              prop_elig * state_next.base.h_hiv_adult(hm, ha, g);
-          intermediate.base.Xartelig_15plus += intermediate.base.artelig_hahm(
-              hm, ha - hIDX_15PLUS);
-          intermediate.base.expect_mort_artelig15plus +=
-              natural_history.cd4_mortality(hm, ha, g) *
-              intermediate.base.artelig_hahm(hm, ha - hIDX_15PLUS);
+
+	  real_type tmp_artelig = prop_elig * state_next.base.h_hiv_adult(hm, ha, g);
```
mwalte10, line 260: Add to the intermediate data object rather than define here.

inst/include/model_simulation.hpp
```diff
@@ -239,87 +239,129 @@ void run_h_art_initiation(int hiv_step,
...
+          intermediate.base.artelig_hahm(hm, ha - hIDX_15PLUS) = tmp_artelig;
+	  intermediate.base.artelig_hm(hm) += tmp_artelig;
+	  intermediate.base.Xartelig_15plus += tmp_artelig;
+
+	  real_type tmp_expect_mort = natural_history.cd4_mortality(hm, ha, g) *
```
mwalte10, line 265: Add to the intermediate data object rather than define here.

inst/include/model_simulation.hpp
```diff
@@ -239,87 +239,129 @@ void run_h_art_initiation(int hiv_step,
...
+      
+      // If the projection period is calendar year (>= Spectrum v6.2), 
+      // this condition is always followed, and it interpolates between
+      // end of last year and current year (+ 1.0).
+      // If projection period was mid-year (<= Spectrum v6.19), the second
+      // half of the projection year interpolates the first half of the
+      // calendar year (e.g. hts 7/10 for 2019 interpolates December 2018
+      // to December 2019)
+
+      real_type art_interp_w = dt * (hiv_step + 1.0);
```
mwalte10, line 330: Add to the intermediate data object rather than define here.

src/frogger.cpp
```diff
@@ -51,10 +51,23 @@ Rcpp::List simulate_model(const leapfrog::StateSpace<ModelVariant> ss,
                           const int proj_years,
                           const int hiv_steps,
                           const std::vector<int> save_steps) {
+
+  int projection_period_int = 0;
+  Rcpp::String projection_period_str  = data["projection_period"];
+  if (projection_period_str  == "midyear") {
+    projection_period_int = 0;
+  } else if (projection_period_str  == "calendar") {
+    projection_period_int = 1;
```
mwalte10, line 60: same suggestion to switch calendar to 0

inst/include/general_demographic_projection.hpp
mwalte10: Add migration_rate_a0 the intermediate data object rather than define here.

inst/include/hiv_demographic_projection.hpp
```diff
@@ -162,19 +161,25 @@ void run_hiv_and_art_stratified_deaths_and_migration(
...
+      }
     }
   }
 
   // remove non-HIV deaths and net migration from adult stratified population
   for (int g = 0; g < ss.NS; ++g) {
     int a = pars.base.options.p_idx_hiv_first_adult;
     for (int ha = 0; ha < ss.hAG; ++ha) {
-      real_type deaths_migrate = 0;
+      real_type deaths_migrate = 0.0;
```
mwalte10, line 177: Add to the intermediate data object rather than define here.

inst/include/hiv_demographic_projection.hpp
```diff
@@ -186,15 +191,61 @@ void run_hiv_and_art_stratified_deaths_and_migration(
...
+    int a = pars.base.options.p_idx_hiv_first_adult;
+    for (int ha = 0; ha < ss.hAG; ++ha) {
+      real_type migration_num_ha = 0.0;
+      real_type hivpop_ha_postmig = 0.0;
+      for (int i = 0; i < ss.hAG_span[ha]; ++i, ++a) {
+	hivpop_ha_postmig += state_next.base.p_hiv_pop(a, g);
+	migration_num_ha += intermediate.base.hiv_net_migration(a, g);
+      }
+      
+      real_type migration_rate = 0.0;
```
mwalte10, line 232: Add to the intermediate data object rather than define here.

inst/include/intermediate_data.hpp
```diff
@@ -102,9 +105,22 @@ struct BaseModelIntermediateData {
...
+  real_type Xart_15plus;
+
+  TensorFixedSize <real_type, Sizes<hDS<ModelVariant>, hAG<ModelVariant>>> artelig_hahm;
+  TensorFixedSize <real_type, Sizes<hDS<ModelVariant>>> artelig_hm;
+  real_type Xartelig_15plus;
+
+  TensorFixedSize <real_type, Sizes<hDS<ModelVariant>>> expect_mort_artelig_hm;
+  real_type expect_mort_artelig15plus;
+
+  TensorFixedSize <real_type, Sizes<hDS<ModelVariant>>> artinit_hm;
```
mwalte10, line 121: Add to the intermediate data object rather than define here.

mwalte10 (changes requested):
See comments on code

src/frogger.cpp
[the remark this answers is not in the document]
jeffeaton: Strongly endorse this

inst/include/model_simulation.hpp
jeffeaton: Regarding this and the other similar comments — my understanding is the `intermediates` object is to avoid re-declaring large temporary blocks of memory on the heap.  This and the other ones flagged are single bits on the stack. They will have negligible overhead.

Is there any benefit to putting these in 'intermediate data'? 

@r-ash maybe weigh in as well.

inst/include/intermediate_data.hpp
jeffeaton: Thanks, this is good suggestion. (They were defined on this based on the sequencing of implementation in EPP-ASM)

jeffeaton: Done in 952c2a8. Also have replaced hardcoding the value to calling the macro

src/frogger.cpp
jeffeaton: In 952c2a8 I have replaced this with calling the macro `leapfrog::internal::PROJPERIOD_CALENDAR`

jeffeaton:
Thanks very much @mwalte10 @r-ash. Can you please review again?

inst/include/model_simulation.hpp
r-ash: Oh yes, I did not realise that. But yes I think no need to add these into intermediate data as they are just stack allocated, sorry if I've misled you on that before Maggie. I think also good to keep that here as keeps the scope of this intermediate data a bit smaller, which is always a positive. I wonder if they are still stack allocated in TMB run where real_type will be one of the TMB types?

jeffeaton: Thanks @r-ash. I'm pretty sure that they will still be for the TMB types, but even if not, it's known at compile time should be well handled.

inst/include/general_demographic_projection.hpp
jeffeaton: Thanks very much, I have elaborated the comment and split up the calculation over two lines.

Surv is the survival probability and the way you have read is correct. In the net migration for all ages above one, half of the mortality is applied `migrant_survival = (0.5 + 0.5 * surv)`. But for infants, the assumption is they will spend more time in the high mortality of the first six months, so 2/3 of the annual mortality is applied.

mwalte10: Sounds good, thanks!

mwalte10 (approved):

PR #44, opened 2024-04-18
Improve how we are managing time steps and update CI

r-ash:
This PR started with just updating the CI. That revealed an issue on valgrind with us reading from uninitialised memory. And in looking at that I think I found a big issue with how we were indexing time. So I have reworked the times here and fixed that.

Before this PR we had an interface to `run_model` where you would set the number of iterations e.g. 60 for the full model. We were then passing this through to C++ and where it became `proj_years` for the projection years and `time_steps` for the number of iterations of the main time loop. `proj_years` was used as one of the dimensions for our input data. But the issue is we were then looping from 1 to <= `time_steps` we were then reading from input data at index `i = time_step` which was reading off the end of any tensor with a `proj_years` index... bit fiddly. We had fixed this in a couple of special cases by passing the input data as `proj_years + 1` but hadn't fixed it everywhere.

Anyway, I thought all that was a bit gross and after chatting through with @mwalte10 she thinks about running the model in terms of years. So I have reverted the interface back to what leapfrog uses which is you call `run_model` with the projection years `1970:2030` by default. We then calculate number of time steps as length of this projection years. I think this gives a cleaner interface, and fixes our valgrind issues.

This PR will
* Updates github actions to our current standard version
* Fixes an instance where we were modifying input data and adds a test to check that this shouldn't happen again
* Adds a script to build test data and runs this
* Updates the `run_model` interface to use projection years instead of time steps

I am expecting the reference tests here will fail, we discussed in the meeting there is some difference between leapfrog and frogger due to updates in leapfrog which we haven't pulled through here. I have skipped these tests for now but will remove that later.


r-ash:
@mwalte10 I've put you as a reviewer on this too, any feedback on the `run_model` interface would be very welcome. I've updated that to work like leapfrog was. So I hope it a bit more intuitive than before.

M-Kusumgar (approved):
Looks good to me!

PR #47, opened 2024-05-29
mrc-5314: Add code generation for state saver types

r-ash:
This PR will add code generation for the OutputState, and StateSaver types. These are the structs which manage the output reporting. I've also moved all the generated code into a directory "generated" within inst/include.

I'd like to do a bit more tidying here, as there are a few things I don't love here
* We still have hardcoded bits in the templates ".hpp.in", I'd like to either generate these, or move them out into another file
* Hardcoded mapping, we have some mapping in the generation code, between ModelVariant and the name used in state e.g. "ModelVariant" to "base", and "ChildModel" to "children".
* Hardcoding of struct types see line 31 in generate_cpp.R
* Specialisations for ChildModel with empty parameters are hardcoded, and will have to be for other new model variants. Would be nice to generate this out. Need to revisit this.
* More directories in inst/include

I think we're a couple of iterations here from something quite tidy and nice but I don't want to make any big structural changes whilst Maggie's PR is in progress as I think will already be a bit of an integration headache.


R/generate_cpp.R
```diff
@@ -412,6 +412,46 @@ generate_state_types <- function(
...
+#' @param output_csv Path to csv file with output specification in it
+#'
+#' @return Nothing, called to generate code in src dir
+#' @keywords internal
+generate_state_saver_types <- function(
+    dest,
+    output_csv = frogger_file("cpp_generation/model_output.csv")) {
+  template_path <- frogger_file("cpp_generation/state_saver_types.hpp.in")
+  template <- readLines(template_path)
+  output_file <- "model_output.csv"
```
M-Kusumgar, line 427: as above?
```suggestion
  output_file <- basename(output_csv)
```

R/generate_cpp.R
```diff
@@ -412,6 +412,46 @@ generate_state_types <- function(
...
 }
 
+#' Generate C++ for state saver types
+#'
+#' @param dest The destination to write generated code to.
+#' @param output_csv Path to csv file with output specification in it
+#'
+#' @return Nothing, called to generate code in src dir
+#' @keywords internal
+generate_state_saver_types <- function(
```
M-Kusumgar, line 422: ive seen the start of this function in a couple of places, mainly cos it's just doing basic parsing, i think this can go into one function, with the header generation as well as these happen for all these generation functions, we should be able to just attach the `header` variable to the parent env inside a function easily, i think the call to the cpp can also be extracted so we are just providing one function that assigns the specific thing that it needs

perhaps will tell you on call!

M-Kusumgar (commented):
Looks good to me! had a small comment about the general structure of this, would be good to discuss on call

R/generate_cpp.R
r-ash: Thanks, that sounds good. I've moved header into the `generate_cpp` function, and updated that to write out the file too. 

I've also put the input and output reading into a util which we can then call from all the generate function. See `read_output_csv` and `read_input_csv`. Let me know what you think

mwalte10 (commented):
This is a bit above my head, but I think I'm tracking where the state_space files are being generated, and that looks good. Flagging that you said there were changes you wanted to make once the paed model was merged in and that now look like its good to go?

PR #49, opened 2024-06-12
Mrc 5466 Run run_wlhiv_births without running full adult model

M-Kusumgar:
I have added a new R function with the same name as the cpp function, `run_wlhiv_births`. This function takes in very similar inputs to the `run_model` function you are used to but with one key difference: it takes in `data_from_adult_model` which is a list of the parameters:
* `h_hiv_adult`
* `h_art_adult`
* `births`
* `p_total_pop`

These were just all Rob and I could find in the `run_wlhiv_births` cpp function. We just substitute these values into `state` and `state_next` at every time step.


man/run_wlhiv_births.Rd
mwalte10, line 1: What do these .Rd files do?

src/frogger.cpp
```diff
@@ -93,3 +93,84 @@ Rcpp::List run_base_model(const Rcpp::List data,
...
+        }
+      }
+      for (int p = 0; p < ss.base.pAG; ++p) {
+        for (int s = 0; s < ss.base.NS; ++s) {
+          state_next.base.p_total_pop(p, s) = p_total_pop(p, s, step);
+          state.base.p_total_pop(p, s) = p_total_pop(p, s, step - 1);
+        }
+      }
+      state_next.base.births = births(step);
+      state.base.births = births(step - 1);
```
mwalte10, line 163: This is just making it so that they state object (not sure if that is the correct word) has the input from the adult population right? Maybe a good thing to create a function for in the future?

man/run_wlhiv_births.Rd
M-Kusumgar: These are R documentation files, you can automatically transform that comment above the run_wlhiv_births R function into documentation automatically, using `devtools::document()` or something like that, i have it binded to a shortcut on my text editor XD

src/frogger.cpp
M-Kusumgar: Yep exactly, this was just a sort of stop gap but I think we have a plan of doing this a more general way which will involve some abstraction of this to help you guys run any general cpp function, that will probably involve either some code generation for this or a smarter function with this logic extracted

r-ash: I have a need for something like this for the Python wrapper. Let me know when you come to impl this and let's chat

R/run_model.R
```diff
@@ -63,3 +63,49 @@ run_model <- function(data, parameters, sim_years,
...
+    output_steps <- which(sim_years %in% output_steps)
+  }
+
+  parameters$hAG_SPAN <- parameters[["hAG_SPAN_full"]]
+  parameters$cd4_initdist <- parameters[["cd4_initdist_full"]]
+  parameters$cd4_prog <- parameters[["cd4_prog_full"]]
+  parameters$cd4_mort <- parameters[["cd4_mort_full"]]
+  parameters$art_mort <- parameters[["art_mort_full"]]
+
+  data <- c(data, parameters)
```
r-ash, line 108: Worth putting this into a helper which we can call from here and run_model?

r-ash (approved):
This looks good, it's a bit gross how much work we have to do to get the setup so we can run the function Maggie needs. I think if this does what Maggie wants, then fine to merge it for now. But we should look at making this easier.

For the Python wrapper I think I am going to need a way to initialise the State from data passed in by Python. I expect we can re-use that bit here to make the State portion a little simpler. 

But I do wonder if worth thinking about pulling `run_wlhiv_births_r` out into another function which just exposes the data Maggie needs. And then calling that from the model code and exposing that function itself. Depends a bit on what that needs.

M-Kusumgar:
closing this for now, this might become obsolete or at least need a lot of modification with the new mixins and code generation pattern

PR #50, opened 2024-06-13
Add Python interface

r-ash:
Note the PR is into an epic feature branch. I don't want this to be a long lived feature branch, but I don't think we should merge this into main until the code generation can be run again

This PR will add a simple python interface which should hopefully act as a starting point. This is implemented as a separate python project within `leapfrog-py` directory. It's a biggie, but a fair bit of this is configuration or quite boring. This PR will
* Add a Python package with 3 functions which I expect to be called externally. I've tried to write these using contract which we want to use for the long term, namely projecting single year will update state in place and C++ will operate on data owned by Python. But I expect these will change a little bit as we actually try and use this from SpectrumEngine. In the following functions, `parameters` and `state` are dicts of numpy arrays. `parameters` are read only but `state` gets written to.
   * set_initial_state - takes `parameters` and `state` and writes the initial state into the `state`, returns nothing
   * project_single_year - takes `i` the time step to run, `parameters` and `state`. Projects a single year forward, writing the result into the `state`
   * run_leapfrog - takes `parameters` and runs the model, by default for years 1970 to 2030 (inclusive) and returns `state` as a dict of numpy arrays
* Add a new python project in `leapfrog-py` directory, I think I might move this eventually but this should be a good starting place. I have checked this on Windows and it is working well there too.
   * This uses uv with [scikit-build-core](https://scikit-build-core.readthedocs.io/en/latest/) as the build backend to compile the C++ code
   * [pybind11](https://pybind11.readthedocs.io/en/stable/) for generating the C++ bindings

Benchmarks for python child model

```
❯ uv run .leapfrog-py/scripts/benchmark.py
Full model benchmark
Minimum Time: 16.0 ms
Median Time: 18.3 ms
Maximum Time: 21.2 ms
Number of Iterations: 50
Total Time: 927.5 ms

Year by year benchmark
Minimum Time: 13.8 ms
Median Time: 16.0 ms
Maximum Time: 20.3 ms
Number of Iterations: 50
Total Time: 839.4 ms
```

Faster than from R

```
Benchmarking child model
# A tibble: 1 × 13
  expression      min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr> <bch:tm> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 frogger      23.8ms 26.3ms      38.3    5.97MB     16.4     7     3      183ms
```

Lots of things still TODO here
* Add CI for Python - *done* #59 
* Add a CI build for pushing to pypi - *done* #126 but I expect this needs revisiting
* Expose the child model instead of the adult model (have just done the child model for now to keep the size of the diff down) - *done* #61 #62 
* Add some test to check that the Python and R code are producing the same result, I expect this might have to live a level above the R or the Python projects - Added a very simple one into Python for now, but we should do this in a better
* Look at why slightly different results from R and Python, I expect that this is difference in the input data. But need to update the input data to confirm that - worked this out, was an input difference
* Expose `State` type which can take some container type. At the moment the `State` is implemented using TensorFixedSize, so I can't pass a reference to numpy data into it and so have to do a copy on the way out. - *done* #60 
   * Could do this by templating it, which is good as that is a compile time switch but means that template then propagates everywhere
   * Or I could do it using std::variant which is like a union type. This is quite nice, but means that it will allocate space for the largest of these every time, which will always be the fixed size tensor. It would be nice to avoid that.

TODO after this
* Settle on packaging structure
* Get code generation working again, including for the Python interface
* Tidy up pypi build and work out when we want to be running this, do we do it on merge into main? Do we do deliberate releases? I expect this has also broken since switching to uv
* Add tests to check that standalone C++, Python, R (and any future wrappers) agree
* Add ability to switch between calendar year and mid year projection into Python interface (hardcoded at the moment)
* Add support for demographic model into Python interface (possibly don't need this immediately)
* Get this integrated into Spectrum Web
* Get documentation written and generated, should this be on the same site as the R code? Or a separate site?
* Generate mypy stubs from pybind11 types https://mypy.readthedocs.io/en/stable/stubgen.html see #58


inst/cpp_generation/model_input.csv
```diff
@@ -12,7 +12,7 @@ cd4_prog,NaturalHistory.cd4_progression,real_type,ModelVariant::run_hiv_simulati
 artcd4elig_idx,Art.idx_hm_elig,int,ModelVariant::run_hiv_simulation,,TRUE,1,proj_years,,,
 cd4_initdist,NaturalHistory.cd4_initial_distribution,real_type,ModelVariant::run_hiv_simulation,,FALSE,3,base.hDS,base.hAG,base.NS,
 art_mort,Art.mortality,real_type,ModelVariant::run_hiv_simulation,,FALSE,4,base.hTS,base.hDS,base.hAG,base.NS
-artmx_timerr,Art.mortaility_time_rate_ratio,real_type,ModelVariant::run_hiv_simulation,,FALSE,2,base.hTS,proj_years,,
+artmx_timerr,Art.mortality_time_rate_ratio,real_type,ModelVariant::run_hiv_simulation,,FALSE,2,base.hTS,proj_years,,
```
r-ash, line 15: Fixing a typo I spotted

Issue #51, opened 2024-06-14
Add CI for Python interface


r-ash:
Run the tests for the first cut, but eventually to build wheels and push to pypyi

PR #59, opened 2024-06-14
Add CI build

r-ash:
Builds on #50 and should be reviewed after that

This PR adds a CI build for python project which runs on windows, linux and mac. The build runs tests, coverage and lint checks. Addresses #51

### DIFF
--- a/.github/workflows/leapfrog-py-test.yaml
+++ b/.github/workflows/leapfrog-py-test.yaml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         config:
           - {os: macos-latest,   py: '3.12'}
-          - {os: ubuntu-latest,  py: '3.8'}
+          - {os: ubuntu-latest,  py: '3.9'}
           - {os: ubuntu-latest,  py: '3.10'}
           - {os: ubuntu-latest,  py: '3.11'}
           - {os: ubuntu-latest,  py: '3.12'}
--- a/.github/workflows/leapfrog-py-test.yaml
+++ b/.github/workflows/leapfrog-py-test.yaml
@@ -1,0 +1,64 @@
+name: Test
+
+on:
+  pull_request:
+    branches:
+      - main
+      - python-interface-poc
+  push:
+    branches:
+      - main
+      - python-interface-poc
+
+jobs:
+  run:
+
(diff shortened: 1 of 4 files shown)


rlglaubius (approved):
Looks good to me, but I rate my confidence as "low" in this space.

.github/workflows/leapfrog-py-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {os: macos-latest,   py: '3.12'}
+          - {os: ubuntu-latest,  py: '3.8'}
+          - {os: ubuntu-latest,  py: '3.10'}
+          - {os: ubuntu-latest,  py: '3.11'}
+          - {os: ubuntu-latest,  py: '3.12'}
+          - {os: windows-latest, py: '3.12'}
```
M-Kusumgar, line 29: why not 3.9?

.github/workflows/leapfrog-py-test.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Test
+        run: |
+          hatch run cov-ci
+
+      - name: Lint
+        run: |
+          # https://github.com/python/mypy/issues/10863
+          mkdir -p .mypy_cache
+          hatch run lint:all
```
M-Kusumgar, line 73: happy with this but might be nicer to run lint before tests, feel like its more immediate feedback

M-Kusumgar (approved):
Looks good to me, have some optional comments!

.github/workflows/leapfrog-py-test.yaml
r-ash: Ah just missed that when I went through and updated the versions. Testing 3.9 - 3.12 which I think should be enough.

.github/workflows/leapfrog-py-test.yaml
r-ash: Yeah makes sense, tests are actually faster than lint atm but will keep in mind as that probably won't always be true

M-Kusumgar: damn tests being faster than lint is amazing

PR #60, opened 2024-06-17
Ensure state types can be passed by reference

### LINKED ISSUE #55: Add a State type which can work with Tensor containers
At the moment the State is implemented using TensorFixedSize, so I can't pass a reference to numpy data into it and so have to do a copy on the way out.
* Could do this by templating it, which is good as that is a compile time switch but means that template then propagates everywhere
* Or I could do it using std::variant which is like a union type. This is quite nice, but means that it will allocate space for the largest of these every time, which will always be the fixed size tensor. It would be nice to avoid that.

r-ash:
Fixes #55 and builds on #59, should be reviewed after that

This PR will
* Add a new template arg to State types so that they can either own data i.e. use TensorFixedSize types, or they can hold references to data owned elsewhere i.e. TensorMap<TensorFixedSize> types. We want this because from Python we have a large numpy array with a time dimension. I want to take a view of this at time `t - 1` and `t`, and pass these in as `state` and `state_next`. Where `state` will be used by reference readonly and `state_next` will be written to. 
* Add a script to benchmark to the python fit

Benchmark with state owning the memory (before this PR)
```
❯ ./benchmark.py
Full model benchmark
Minimum Time: 16.5 ms
Median Time: 19.3 ms
Maximum Time: 24.3 ms
Number of Iterations: 50
Total Time: 980.6 ms

Year by year benchmark
Minimum Time: 18.6 ms
Median Time: 19.4 ms
Maximum Time: 22.9 ms
Number of Iterations: 50
Total Time: 987.2 ms
```

Passing State by reference (after this PR)
```
❯ ./scripts/benchmark.py
Full model benchmark
Minimum Time: 16.2 ms
Median Time: 19.5 ms
Maximum Time: 22.8 ms
Number of Iterations: 50
Total Time: 980.5 ms

Year by year benchmark
Minimum Time: 15.2 ms
Median Time: 15.7 ms
Maximum Time: 17.6 ms
Number of Iterations: 50
Total Time: 799.2 ms
```

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -2,9 +2,7 @@
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   push:
-    branches: [ main, master ]
   pull_request:
-    branches: [ main, master ]
 
 name: R-CMD-check
 
--- a/.github/workflows/leapfrog-py-test.yaml
+++ b/.github/workflows/leapfrog-py-test.yaml
@@ -2,13 +2,7 @@ name: Test
 
 on:
   pull_request:
-    branches:
-      - main
-      - python-interface-poc
   push:
-    branches:
-      - main
-      - python-interface-poc
 
 jobs:
   run:
--- a/inst/include/child_model_simulation.hpp
+++ b/inst/include/child_model_simulation.hpp
@@ -7,11 +7,11 @@ namespace leapfrog {
 
 namespace internal {
 
-template<typename ModelVariant, typename real_type>
+template<typename ModelVariant, typename real_type, bool OwnedData>
 void run_child_ageing(int time_step,
                       const Parameters<ModelVariant, real_type> &pars,
-                      const State<ModelVariant, real_type> &state_curr,
-                      State<ModelVariant, real_type> &state_next,
+                      const State<ModelVariant, real_type, OwnedData> &state_curr,
+                      State<ModelVariant, real_type, OwnedData> &state_next,
                       IntermediateData<ModelVariant, real_type> &intermediate) {
   const auto demog = pars.base.demography;
   const auto cpars = pars.children.children;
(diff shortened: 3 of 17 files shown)


rlglaubius (approved):
Looks good!

M-Kusumgar (commented):
I think the code looks good, however, was wondering how much the different tensor types actually affect run time, have you benchmarked the difference (for the R)? If the difference isnt very big/negligible then it might be worth it to not have yet another template arg and instead just treat everything as a TensorMap of TensorFixedSize. I believe promoting to a tensor map is probably very cheap

r-ash:
> I think the code looks good, however, was wondering how much the different tensor types actually affect run time, have you benchmarked the difference (for the R)? If the difference isnt very big/negligible then it might be worth it to not have yet another template arg and instead just treat everything as a TensorMap of TensorFixedSize. I believe promoting to a tensor map is probably very cheap

I'm not planning to use the mapped tensor types from R, and would require some changes to get it to do that. We want `State` to always own that data. And I *think* though not 100% sure that that would require some more changes than just wrapping the TensorFixedSize in a TensorMap.

If I understand what you're suggesting right, something like this? 

```
struct State {
  TensorMap<TensorFixedSize<double, Sizes<3, 4>>> a;

  State() : a(TensorFixedSize<double, Sizes<3, 4>>().data(), Sizes<3, 4>()) {}

  State(const TensorMap<TensorFixedSize<double, Sizes<3, 4>>>& a) : a(a) {}

  void reset() {
    a.setZero();
  }
};
```

I think with this we'll run into an issue with the TensorFixedSize being temporary. So believe after calling the no arg constructor the TensorFixedSize will be temporary and so will be cleaned up. Leaving the TensorMap with a dangling pointer. So I believe for this to work we'd either need to create those TensorFixedSizes somewhere else and ensure they live longer than the State object. Or we'd need State to hold a reference to their data. e.g.

```
TensorFixedSize<double, Sizes<3, 4>> a_data;
TensorMap<TensorFixedSize<double, Sizes<3, 4>>> a;
```

which is a bit gross.

So I think using the template param is perhaps the "nicest" way to do this. I also use it in a later PR with `std::enable_if_t` to exclude certain functions from the compiled version. Which I think is quite nice. Stops accidental calls to wrong ctor or to `reset` when it shouldn't be allowed. See https://github.com/mrc-ide/frogger/pull/62/files#diff-3180c63332b3fc5017131b639495e7e3eaadf9c26cad780d96ba2990eeef6bd1R49

That said - I'm just building understanding of this atm. And very happy to be corrected if I am misunderstanding something.

M-Kusumgar (approved):
Think I understand this now! Happy to approve it!

Issue #52, opened 2024-06-14
Expose the child model from the Python bindings


r-ash:
This is what we will want to run eventually, but just exposed the adult model for the time being as we get things running.

Will need to
* Add test data
* Update the bindings

PR #61, opened 2024-06-18
Tidy up the test test scripts in preparation for adding the child model

r-ash:
This PR build on #60 and should be reviewed after that. This sets up the and tidies the test data we need for #52 but doesn't expose the child model yet. That will be in the next PR

This PR will
* Update the test data and ensure standalone test data is consistent with the other test data. If you run `./scripts/create_test_data.R` or `./scripts/update_standalone_data` after this PR it shouldn't update any of the existing files
* Store the standalone test data in github in a compressed format, wanted to do this so when we add the data for testing child model it isn't quite as big in github

Started pulling this thread when I was thinking about how to test the child model via python. With the current adult model I am using the standalone data. This PR is about updating that standalone test data, improving how we are storing it a bit. Fixing the inconsistent standalone and test data on disk.


scripts/read_spectrum.R
```diff
@@ -290,8 +290,8 @@ prepare_hc_leapfrog_projp <- function(pjnz, params, pop_1){
 
   #cd4
   wlhiv_cd4 <- array(as.numeric(unlist(dpsub("<CD4Distribution15_49 MV2>", 19:25, timedat.idx))), dim = c(7,length(timedat.idx)))
-  v$prop_gte350 <- colSums(wlhiv_cd4[1:2,]) / colSums(wlhiv_cd4)
-  v$prop_lt200 <- colSums(wlhiv_cd4[5:7,]) / colSums(wlhiv_cd4)
+  v$prop_gte350 <- ifelse(colSums(wlhiv_cd4) == 0, 0, colSums(wlhiv_cd4[1:2,]) / colSums(wlhiv_cd4))
+  v$prop_lt200 <- ifelse(colSums(wlhiv_cd4) == 0, 0, colSums(wlhiv_cd4[5:7,]) / colSums(wlhiv_cd4))
```
r-ash, line 294: This was producing NaNs for first few entries, have set these to 0 instead

M-Kusumgar (commented):
Code looks good but when I ran the `create_test_data.R` script the `child_parms.rds` in the test data showed up as modified, im assuming this shouldnt be the case?

r-ash:
> Code looks good but when I ran the `create_test_data.R` script the `child_parms.rds` in the test data showed up as modified, im assuming this shouldnt be the case?
> 
> Can confirm the child parms dont change after the first time I run the script so perhaps you just forgot to commit that?

Oh that's a bit weird, it doesn't for me? Definitely the `rds` file? Not the zip? And after a fresh install?

PR #62, opened 2024-06-21
Expose the child model via python bindings

### LINKED ISSUE #52: Expose the child model from the Python bindings

r-ash:
This PR exposes the child model via the python interface. Closes #52 

Couple of things to note:
* I also tidy a bit of the `OwnedData` interface so only specific ctors are enabled in the compiled code depending on if `OwnedData` is true or false.
* I've updated all the State types to be Eigen::Tensors or Eigen::TensorMaps. We want to be able to pass a reference type for the State when running via Python. But we had a couple of state members which were just a `real_type`. I've updated these to an eigen type so we can pass the container type. Note that when using them you'll need to select the first element `(0)`
* There is still a difference between the model in R and python. I'm not sure where this is coming from as I think the data is the same. I'll dig into this at a later point.


scripts/read_spectrum.R
```diff
@@ -112,8 +112,7 @@ prepare_hc_leapfrog_projp <- function(pjnz, params, pop_1){
 
   ## pull in ART coverage numbers
   art = input_childart(pjnz)
-  v$artpaeds_isperc <- art$art_ispercent[2,]
-  v$artpaeds_isperc[] <- as.integer(ifelse(art$art_ispercent[2,] == FALSE, 0, 1))
+  v$artpaeds_isperc <- as.integer(vapply(art$art_ispercent[2,], isTRUE, logical(1)))
```
r-ash, line 115: This was previously introducing NAs. Note:

```
> NA == FALSE
[1] NA
```

You can use `isFALSE`, `isTRUE` for this 

```
> isTRUE(NA)
[1] FALSE
> isTRUE(FALSE)
[1] FALSE
> isTRUE(TRUE)
[1] TRUE
> isTRUE(c(1, 2, 3))
[1] FALSE
> isTRUE(1)
[1] FALSE
```

M-Kusumgar (approved):
LGTM, i like the simplification of using a tensor to represent a real honestly!

scripts/read_spectrum.R
rlglaubius: I'm not sure why art_ispercent would have NAs in it "ChildARTByAgeGroupPerNum MV2" should have five rows (cotrim, 0-14 ART, 0-4 ART, 5-9 ART, 10-14 ART) with values 0 or 1 by year. Is it possible the NAs are introduced when reading an older Spectrum file ("MV" instead of "MV2"?). If so, we probably need to address that instead of crushing NA's at this stage of processing.

r-ash: Thanks! Good spot, so we were reading from the correct tag but NAs were being introduced as we had mixed types. First few columns were characters and next were numeric. Then we were trying to do a as.logical on this column by column which doesn't work for character types.

```
as.logical("0")
[1] NA
```

I've updated this so actually addressing the root cause instead of patching it up here.

I've updated to using type-safe vapply too instead of sapply. 

I've re-run the `./scripts/create_test_data.R` script with this new code and nothing has changed in the rds. So I think this has sorted it.

Issue #64, opened 2024-06-21
Add some CI check that we're not reading out of bounds of tensors


r-ash:
I feel like we have run into this a few times, and added the valgrind CI build in hopes it would capture this kind of thing. But valgrind won't necessarily capture reading out of bounds.

Scenario here, we have e.g. a 10x10 Eigen::Tensor, but we index in at (11, 11). That might
1. Read from unallocated memory - valgrind will tell us about this
2. Access another variable's memory if it is at that location. Valgind might not necessarily tell us about this.

So we need a better way to do this. Turns out Eigen has that built in, just not well documented via `EIGEN_INTERNAL_DEBUGGING` preprocessor directive. If we compile our code with `-DEIGEN_INTERNAL_DEBUGGING` it will throw a SIGSEGV. e.g.

```
> out <- run_model(demp, parameters, 1970:2030, NULL, run_child_model = TRUE)
R: /home/rob/.R/lib/RcppEigen/include/unsupported/Eigen/CXX11/src/Tensor/TensorMap.h:164: Eigen::TensorMap<PlainObjectType, Options_, MakePointer_>::StorageRefType Eigen::TensorMap<PlainObjectType, Options_, MakePointer_>::operator()(Eigen::TensorMap<PlainObjectType, Options_, MakePointer_>::Index) const [with PlainObjectType = Eigen::Tensor<double, 1, 0, long int>; int Options_ = 0; MakePointer_ = Eigen::MakePointer; Eigen::TensorMap<PlainObjectType, Options_, MakePointer_>::StorageRefType = double&; Eigen::TensorMap<PlainObjectType, Options_, MakePointer_>::Index = long int]: Assertion `index >= 0 && index < size()' failed.
Aborted (core dumped)
```

This is great, but we really need a stack trace with this to make it useful. Can we use `backtrace` with this to detect issues with reading out of bounds? https://stackoverflow.com/questions/77005/how-to-automatically-generate-a-stacktrace-when-my-program-crashes

Set up some CI build which will compile in debug mode, run the child model and then capture a backtrace on any failures

r-ash:
After switch to ndarray will this still be relevant? I think that might check out of the box?

r-ash:
> After switch to ndarray will this still be relevant? I think that might check out of the box?

Sadly, it does not. Is there anything else we can do?

PR #65, opened 2024-06-21
Fix read out of bounds

r-ash:
This PR fixes an instance I have seen when adding the Python bindings where we are reading out of bounds of one of the tensors.

We've had this come up a couple of times and the valgrind build isn't quite addressing it. See #64 for an idea of how we can get some proper checks for this. For now, this just fixes the problem. @mwalte10 please let me know if fixes I've made are appropriate or if there is a better way to handle this.

### DIFF
--- a/inst/include/child_model_simulation.hpp
+++ b/inst/include/child_model_simulation.hpp
@@ -1383,12 +1383,10 @@ void fill_model_outputs(int time_step,
         }// end hcTT
 
         for (int dur = 0; dur < ss.hTS; ++dur) {
-          if (state_next.children.hc1_art_pop(0, hd, a, s) > 0) {
-            if (a < hc_ss.hc2_agestart) {
-              state_next.base.p_hiv_deaths(a,s) += state_next.children.hc1_art_aids_deaths(dur, hd, a, s);
-            } else if (hd < (hc_ss.hc2DS)) {
-              state_next.base.p_hiv_deaths(a,s) += state_next.children.hc2_art_aids_deaths(dur, hd, a-hc_ss.hc2_agestart, s);
-            }
+          if (a < hc_ss.hc2_agestart && state_next.children.hc1_art_pop(0, hd, a, s) > 0) {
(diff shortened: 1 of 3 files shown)


inst/include/child_model_simulation.hpp
```diff
@@ -1383,12 +1383,10 @@ void fill_model_outputs(int time_step,
         }// end hcTT
 
         for (int dur = 0; dur < ss.hTS; ++dur) {
-          if (state_next.children.hc1_art_pop(0, hd, a, s) > 0) {
```
r-ash, line 1386: The issue here is `a`. The loop run `s` up to `pars.base.options.p_idx_fertility_first` which is 15. But `hc1_art_pop`'s 3rd dimension is age which runs up to `hc1AG` which is `5` so we were reading off the end of the data here.

inst/include/child_model_simulation.hpp
```diff
@@ -1383,12 +1383,10 @@ void fill_model_outputs(int time_step,
...
-            if (a < hc_ss.hc2_agestart) {
-              state_next.base.p_hiv_deaths(a,s) += state_next.children.hc1_art_aids_deaths(dur, hd, a, s);
-            } else if (hd < (hc_ss.hc2DS)) {
-              state_next.base.p_hiv_deaths(a,s) += state_next.children.hc2_art_aids_deaths(dur, hd, a-hc_ss.hc2_agestart, s);
-            }
+          if (a < hc_ss.hc2_agestart && state_next.children.hc1_art_pop(0, hd, a, s) > 0) {
+            state_next.base.p_hiv_deaths(a,s) += state_next.children.hc1_art_aids_deaths(dur, hd, a, s);
+          } else if (hd < (hc_ss.hc2DS) && a >= hc_ss.hc2_agestart) {
+            state_next.base.p_hiv_deaths(a,s) += state_next.children.hc2_art_aids_deaths(dur, hd, a-hc_ss.hc2_agestart, s);
```
r-ash, line 1389: Similar to above, `a` is the problem here, if we had `a` < `hc_ss.hc2_agestart` then this was using a -ve index

mwalte10 (approved):

PR #68, opened 2024-06-26
Add test to compare R and Python model output

r-ash:


R/generate_cpp.R
```diff
@@ -609,3 +609,34 @@ read_types_csv <- function(path, validation_function, group_by_col) {
...
+
+
+#' Get mapping of R names to C++ names for input data
+#'
+#' @param input_csv Path to input csv which is source of truth for these names
+#'
+#' @return A list with r_name, cpp_name, and C++ struct/namespace this
+#'   belongs to
+#' @export
+frogger_input_name_mapping <- function(
```
r-ash, line 621: The noisiest thing here is that I have added this function. I was getting a bit annoyed with having multiple mappings of R data to C++ data from the scripts. I think I had just avoided writing that previously as I was being lazy. We should now use this anywhere we need to map between the two names. I'm using this in 2 places atm
1. In script to create the python and C++ standalone test data to map from saved R names into C++ names on disk
2. In script to run the R interface to the model (which is used for testing Python & R produce the same result) to map from the C++ names on disk, back into R names

R/run_model.R
```diff
@@ -47,13 +47,11 @@ run_model <- function(data, parameters, sim_years,
     stop("Cannot run child model with coarse age stratification")
   }
   if (hiv_age_stratification == "full") {
-    parameters$hAG_SPAN <- parameters[["hAG_SPAN_full"]]
     parameters$cd4_initdist <- parameters[["cd4_initdist_full"]]
     parameters$cd4_prog <- parameters[["cd4_prog_full"]]
     parameters$cd4_mort <- parameters[["cd4_mort_full"]]
     parameters$art_mort <- parameters[["art_mort_full"]]
   } else {
-    parameters$hAG_SPAN <- parameters[["hAG_SPAN_coarse"]]
```
r-ash, line 56: I've removed these as they weren't actually being used, this is in the state space

R/util.R
```diff
@@ -42,8 +42,12 @@ serialize_r_to_tensor <- function(data, path) {
 #' @keywords internal
 deserialize_tensor_to_r <- function(path) {
   content <- readLines(path)
+  converter <- switch(
+    content[[1]],
+    "int" = as.integer,
```
r-ash, line 47: Turns out our test for this wasn't actually checking the type! See https://github.com/mrc-ide/frogger/pull/68/files#diff-9019d605dcb8ef5d76c288c13a1c7483f042c215ef2e49b4d4f0b932cad296eeR81 Difference between `expect_equal` and expect_identical.

rlglaubius (approved):
I don't have a strong opinion about the code changes, but I'm glad to have a standard way of cross-checking outputs between Python and R.

PR #73, opened 2024-07-05
extract out cpp functions into R - TMB epic base

### LINKED ISSUE #69: Extract transform function at start of run_base_model into R code
This code can easily be put into R, this is in preparation for adding TMB into frogger. TMB requires a different way to get inputs from R, which means we would need duplicates of each of the transform functions for TMB inputs and Rcpp inputs. Instead we can just do this in one place in R

M-Kusumgar:
This resolves #69 and also serves as the base for the TMB epic. We were transforming some function args in cpp for our code but we can just do this in R. If we had kept it in cpp then we would've had to include copies of the functions for TMB so porting them to R seemed the easiest.


src/frogger.cpp
```diff
@@ -8,43 +9,6 @@
...
 
-int transform_simulation_years(
-    const Rcpp::List demp,
-    Rcpp::Nullable<Rcpp::NumericVector> r_sim_years) {
-
-  Rcpp::NumericVector Sx = demp["Sx"];
-  Rcpp::Dimension d = Sx.attr("dim");
-  // Simulation initialises state from first years input data (index 0)
-  // then runs for each year simulating this years (i) data using previous years
-  // state (i - 1) and this years input data (i).
```
r-ash, line 19: Can we keep this note somewhere? I think this is v easy to forget but quite important for making sense of the looping

R/run_model.R
```diff
@@ -61,5 +61,34 @@ run_model <- function(data, parameters, sim_years,
...
+  hiv_steps <- transform_hts_per_year(hts_per_year)
+
+  run_base_model(data, proj_years, hiv_steps, save_steps, model_variant)
+}
+
+transform_simulation_years <- function(data, sim_years) {
+  sx <- data$Sx
+  max_sim_years <- dim(sx)[3]
+  if (is.null(sim_years)) {
+    return(max_sim_years)
```
r-ash, line 76: Probably worth adding small test for this

R/run_model.R
```diff
@@ -61,5 +61,34 @@ run_model <- function(data, parameters, sim_years,
...
+  }
+  n_years <- length(sim_years)
+  if (n_years > max_sim_years) {
+    stop(paste0("No of years > max years of ", max_sim_years))
+  }
+  n_years
+}
+
+transform_output_steps <- function(output_steps) {
+  output_steps - 1
```
r-ash, line 86: Maybe worth adding a note on this one to make it clear we're doing this because of 1-based indexing vs 0-based indexing

r-ash (commented):
This looks great, small comments but looks much simpler to put those transformation functions into R

M-Kusumgar:
will do in new repo

PR #74, opened 2024-07-05
Separate out TMB from rcpp export and switch on model type

### LINKED ISSUE #70: Add model switching logic TMB
this should be a small PR hopefully but just add the initial model switching logic by getting the data string in TMB and switching on that

M-Kusumgar:
This resolves #70. Main points:
* new cpp file `frogger_TMB.cpp` for all the TMB related things. The reason I had to do this is because the `frogger` DLL was being created by Rcpp and it was registering **only** the routines we created and not TMB's. We need to create a separate DLL for TMB which doesn't use any Rcpp so that all the routines are exported. There is a way around this, and i know how to do it but Rich does not think it is a good idea, he's probably right about it. Basically involves us writing our own version of what RcppExports.cpp file looks like and adding the TMB routines into the `CallEntries` array (TMB exports all the routines it needs) so it would be a one line change to that file. But yh we would be the ones that have to manage that.
* had to add a `compile.R` script that gets run via the Makevars (Makevars.win for windows systems), this code was taken from a package that uses TMB and is on CRAN, https://github.com/tsmodels/tsgarch/blob/main/src/TMB/compile.R, Makevars were similarly taken from that package too. This is seen in multiple packages so probably an established pattern of how to integrate Rcpp and TMB.
* added a temporary `test_tmb` R function because it would be nice to actually keep checking that things actually work while we develop this.
* added a simple test setup, `parameters$incidinput` was not a list for some reason, it is the only exception so i just made it a list, let me know if this is a problem or not equivalent. Currently TMB will just return a 0 and output some mgc stuff so if you want to print something from TMB function to test it out, use `Rcout` macro like so:
```R
  Rcout << model_variant;
```
it works like normal `cout` but for TMB. It will print 4 times though because the template is run 4 times


R/run_model.R
```diff
@@ -92,3 +92,73 @@ transform_hts_per_year <- function(hts_per_year) {
...
+    parameters$cd4_prog <- parameters[["cd4_prog_coarse"]]
+    parameters$cd4_mort <- parameters[["cd4_mort_coarse"]]
+    parameters$art_mort <- parameters[["art_mort_coarse"]]
+  }
+  data <- c(data, parameters)
+
+
+  proj_years <- transform_simulation_years(data, sim_years)
+  save_steps <- transform_output_steps(output_steps)
+  hiv_steps <- transform_hts_per_year(hts_per_year)
```
r-ash, line 149: Want to put this into a helper function to remove the duplication with `run_model`?

R/run_model.R
```diff
@@ -92,3 +92,73 @@ transform_hts_per_year <- function(hts_per_year) {
...
+    model_variant = model_variant,
+    proj_years = proj_years,
+    hiv_steps = hiv_steps,
+    save_steps = save_steps
+  )
+  data_tmb <- c(data, extra_inputs)
+  parameter_tmb <- list(x = 0)
+  obj <- TMB::MakeADFun(data_tmb, parameter_tmb, DLL = "frogger_TMB")
+  obj$hessian <- TRUE
+  opt <- do.call("optim", obj)
```
r-ash, line 163:
```suggestion
  do.call("optim", obj)
```

r-ash (commented):
Looks good, but yeah would be useful to chat through the TMB stuff. I can't remember
* When does the TMB code get compiled, that'll be when you load the package?

r-ash (approved):

M-Kusumgar:
will do in new repo

src/TMB/frogger_TMB.cpp
```diff
@@ -0,0 +1,54 @@
+#include <Rcpp.h>
+#include <TMB.hpp>
+
+#include "frogger.hpp"
+#include "intermediate_data.hpp"
+#include "model_variants.hpp"
+#include "state_space.hpp"
+
+#undef TMB_OBJECTIVE_PTR
+#define TMB_OBJECTIVE_PTR obj
```
r-ash, line 10: Required for putting TMB code in different files

PR #107, opened 2024-09-13
improve code readability example

M-Kusumgar:
This draft PR is just a demo of the readability of code improvement. The github diff does not do a great job of displaying this would be better if you check out to this branch in your editor.

Let me know your thoughts!

### DIFF
--- a/inst/include/child_model_simulation.hpp
+++ b/inst/include/child_model_simulation.hpp
@@ -13,56 +13,56 @@ void run_child_ageing(int ts,
                       const State<ModelVariant, real_type> &state_curr,
                       State<ModelVariant, real_type> &state_next,
                       IntermediateData<ModelVariant, real_type> &intermediate) {
-  const auto& dmp = pars.base.demography;
-  const auto& hcp = pars.children.children;
-  const auto& opp = pars.base.options;
-  const auto& hcc = state_curr.children;
-  const auto& hcn = state_next.children;
+  const auto& dm_p = pars.base.demography;
+  const auto& hc_p = pars.children.children;
+  const auto& op_p = pars.base.options;
+  const auto& hc_c = state_curr.children;
+  const auto& hc_n = state_next.children;
 
   static_assert(ModelVariant::run_child_model,
                 "run_hiv_child_infections can only be called for model variants where run_child_model is true");
   constexpr StateSpace<ModelVariant> ss = StateSpace<ModelVariant>();
-  constexpr auto ssb = ss.base;
-  constexpr auto ssc = ss.children;
+  constexpr auto ss_b = ss.base;
+  constexpr auto ss_c = ss.children;
 
-  for (int s = 0; s < ssb.NS; ++s) {
+  for (int s = 0; s < ss_b.NS; ++s) {
     //less than 5 because there is a cd4 transition between ages 4 and 5
-    for (int a = 1; a < ssc.hc2_agestart; ++a) {
-      for (int hd = 0; hd < ssc.hc1DS; ++hd) {
-        for (int cat = 0; cat < ssc.hcTT; ++cat) {
-          hcn.hc1_hiv_pop(hd, cat, a, s) += hcc.hc1_hiv_pop(hd, cat, a - 1, s) * dmp.survival_probability(a, s, ts);
+    for (int a = 1; a < ss_c.hc2_agestart; ++a) {
+      for (int hd = 0; hd < ss_c.hc1DS; ++hd) {
+        for (int cat = 0; cat < ss_c.hcTT; ++cat) {
+          hc_n.hc1_hiv_pop(hd, cat, a, s) += hc_c.hc1_hiv_pop(hd, cat, a - 1, s) * dm_p.survival_probability(a, s, ts);
         }
-        for (int dur = 0; dur < ssb.hTS; ++dur) {
-          hcn.hc1_art_pop(dur, hd, a, s) += hcc.hc1_art_pop(dur, hd, a - 1, s) * dmp.survival_probability(a, s, ts);
-        }
-      }
-    }
-  }
-
-  for (int s = 0; s < ssb.NS; ++s) {
-    for (int hd = 0; hd < ssc.hc1DS; ++hd) {
-      for (int hd_alt = 0; hd_alt < ssc.hc2DS; ++hd_alt) {
-        const auto cd4_transition_prob = dmp.survival_probability(ssc.hc2_agestart, s, ts) * hcp.hc_cd4_transition(hd_alt, hd);
-        for (int cat = 0; cat < ssc.hcTT; ++cat) {
-          hcn.hc2_hiv_pop(hd_alt, cat, 0, s) += hcc.hc1_hiv_pop(hd, cat, ssc.hc1_ageend, s) * cd4_transition_prob;
-        }
-        for (int dur = 0; dur < ssb.hTS; ++dur) {
-          hcn.hc2_art_pop(dur, hd_alt, 0, s) += hcc.hc1_art_pop(dur, hd, ssc.hc1_ageend, s) * cd4_transition_prob;
+        for (int dur = 0; dur < ss_b.hTS; ++dur) {
+          hc_n.hc1_art_pop(dur, hd, a, s) += hc_c.hc1_art_pop(dur, hd, a - 1, s) * dm_p.survival_probability(a, s, ts);
         }
       }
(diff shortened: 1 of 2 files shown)


inst/include/child_model_simulation.hpp
```diff
@@ -13,54 +13,70 @@ void run_child_ageing(int time_step,
...
+
+  const auto hc1_hiv_pop_curr = state_curr.children.hc1_hiv_pop;
+  const auto hc1_art_pop_curr = state_curr.children.hc1_art_pop;
+  const auto hc2_hiv_pop_curr = state_curr.children.hc2_hiv_pop;
+  const auto hc2_art_pop_curr = state_curr.children.hc2_art_pop;
+
+  auto hc1_hiv_pop_next = state_next.children.hc1_hiv_pop;
+  auto hc1_art_pop_next = state_next.children.hc1_art_pop;
+  auto hc2_hiv_pop_next = state_next.children.hc2_hiv_pop;
+  auto hc2_art_pop_next = state_next.children.hc2_art_pop;
```
jeffeaton, line 28: This is fine to me. But also, I think that the previous structure here with shortening `pars.base.demography` -> `demog` and `pars.children.children` -> `cpars` is pretty good and perhaps preferred.

That is in contrast to, for example, the below where `intermediate.base.*****` gets written out a lot and `pars.base.natural_history.*****` gets written out, even though there is a prefix:
https://github.com/mrc-ide/frogger/blob/bd7ee9b00e1de1ed44f9aff28fffc83a4336b046/inst/include/model_simulation.hpp#L56-L76


I might advocate for further standardizing these: e.g. `demog` -> `dempar` and `cpars` -> `chpar` or something like that

And `natural_history.*****` in the adult model, prefer shortening to something like `nh.*****`

(Maybe we standardise them all to 2-letter prefixes and just make some notes about the conventions there)

inst/include/child_model_simulation.hpp
```diff
@@ -13,54 +13,70 @@ void run_child_ageing(int time_step,
...
+
+  for (int s = 0; s < NS; ++s) {
     //less than 5 because there is a cd4 transition between ages 4 and 5
-    for (int a = 1; a < hc_ss.hc2_agestart; ++a) {
-      for (int hd = 0; hd < hc_ss.hc1DS; ++hd) {
-        for (int cat = 0; cat < hc_ss.hcTT; ++cat) {
-          state_next.children.hc1_hiv_pop(hd, cat, a, s) +=
-            state_curr.children.hc1_hiv_pop(hd, cat, a - 1, s) * demog.survival_probability(a, s, time_step);
+    for (int a = 1; a < hc2_agestart; ++a) {
+      const auto survival_prob_per_group = survival_probability(a, s, time_step);
```
jeffeaton, line 44: I don't like this change; prefer keeping the indexes referenced in the line below to easily parse what the calculation is.

i.e., I prefer: 
```
hc1_hiv_pop_next(hd, cat, a, s) += hc1_hiv_pop_curr(hd, cat, a - 1, s) * survival_probability(a, s, time_step)
```
instead of
```
hc1_hiv_pop_next(hd, cat, a, s) += hc1_hiv_pop_curr(hd, cat, a - 1, s) * survival_prob_per_group
```

If we want to make these more concise for readability, I would be open to abbreviating `time_step` to something shorter.

inst/include/child_model_simulation.hpp
```diff
@@ -13,54 +13,70 @@ void run_child_ageing(int time_step,
...
-    for (int a = (hc_ss.hc2_agestart + 1); a < pars.base.options.p_idx_fertility_first; ++a) {
-      for (int hd = 0; hd < hc_ss.hc2DS; ++hd) {
-        for (int cat = 0; cat < hc_ss.hcTT; ++cat) {
-          state_next.children.hc2_hiv_pop(hd, cat, a - hc_ss.hc2_agestart, s) +=
-            state_curr.children.hc2_hiv_pop(hd, cat, a - hc_ss.hc2_agestart - 1, s) *
-            demog.survival_probability(a, s, time_step);
+  for (int s = 0; s < NS; ++s) {
+    for (int a = (hc2_agestart + 1); a < p_idx_fertility_first; ++a) {
+      const auto survival_prob_per_group = survival_probability(a, s, time_step);
+      const auto hc2_age = a - hc2_agestart;
```
jeffeaton, line 73: I am take or leave on this one.

But if we are going to change, would like the name `hc2_age` to indicate that this is referring to an array index (rather than an age)

jeffeaton (commented):
Thanks @M-Kusumgar for the example on this. I've left some comments — we can talk through or discuss on here.

inst/include/child_model_simulation.hpp
r-ash: I agree with this. Switching `time_step` to `ts` or similar in the function signature sounds sensible.

inst/include/child_model_simulation.hpp
r-ash: So suggesting Jeff that e.g. instead of the line

```
hc1_hiv_pop_next(hd, cat, a, s) += hc1_hiv_pop_curr(hd, cat, a - 1, s) * survival_probability(a, s, time_step);
```

You would have?

```
cpars.hc1_hiv_pop_next(hd, cat, a, s) += cpars.hc1_hiv_pop_curr(hd, cat, a - 1, s) * demog.survival_probability(a, s, time_step);
```

That is less unpacking! So definitely a fan of that if it still seems readable to you? (I think the first is easier to read)

inst/include/child_model_simulation.hpp
```diff
@@ -13,54 +13,70 @@ void run_child_ageing(int time_step,
...
+  const auto hc_cd4_transition = pars.children.children.hc_cd4_transition;
+  const auto p_idx_fertility_first = pars.base.options.p_idx_fertility_first;
+
+  const auto hc1_hiv_pop_curr = state_curr.children.hc1_hiv_pop;
+  const auto hc1_art_pop_curr = state_curr.children.hc1_art_pop;
+  const auto hc2_hiv_pop_curr = state_curr.children.hc2_hiv_pop;
+  const auto hc2_art_pop_curr = state_curr.children.hc2_art_pop;
+
+  auto hc1_hiv_pop_next = state_next.children.hc1_hiv_pop;
+  auto hc1_art_pop_next = state_next.children.hc1_art_pop;
```
r-ash, line 26: I think we want to use `const auto&` and `auto&` here to ensure that these don't copy

r-ash (commented):
This looks good! It does look much more readable with these changes

inst/include/child_model_simulation.hpp
jeffeaton: Thanks, I am happy with the first also, it just seemed like a lot of lines of assignment transcribing at the top of the function. 

Would you be open to shortening the prefixes + what do you think, e.g.

```
auto hc = state.children;
auto dp = pars.base.demography;
hc.hc1_hiv_pop_next(hd, cat, a, s) += hc.hc1_hiv_pop_curr(hd, cat, a - 1, s) * dp.survival_probability(a, s, time_step)
```

(I think `hc1_hiv_pop_next` would not belong to `cpars.`)

M-Kusumgar: rob and I also had a chat and he agrees that the prefixes are fine! so will be doing that, ill put together a list of proposed prefix names for the meeting tomorrow

inst/include/child_model_simulation.hpp
M-Kusumgar: sounds good, then for now i wont be introducing any proxy variables (would you like proxy variables for multiple tensor expression, for example [this](https://github.com/mrc-ide/frogger/pull/107/files#diff-563821e9ce79549a4ac984cd4b62632d6659a4dc04fbc27aebd9c52b414477d1R59), can talk about it in the meeting), also will be abbreviating `time_step` to `ts`

inst/include/child_model_simulation.hpp
M-Kusumgar: im going to opt to not do this then because we are reducing the amount of proxy variables anyway and shortening `time_step`, i think itll be terse enough but if it does become an issue after the rewrite, i can always just do another PR that addresses all of these instances

M-Kusumgar:
as discussed, can we also also add all the rules about naming into the vignettes

M-Kusumgar:
this can be left for documenting the discussion but closing as first parts of the refactor are already in

PR #129, opened 2025-02-04
Add separate model variant for demographic projection

### LINKED ISSUE #91: Add model variant to run DP as a standalone fit
We need this in avenir, two ways of running the model
1. DP projection only
2. DP + AIM projection

Add a way to do this switching, probably with a model variant.

r-ash:
This PR will
* Create a new model variant `DemographicProjection` 

Closes #91 

This has created a bit of noise! Most important for review I think is I had to update the structure of the parameters and how we pass these in in `model_input.hpp`. I had punted on that before as we just had a single conditional but now there is a new variant that approach needed some refinement.

This still needs a bit of tidying before we can merge it


configure
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-$R_HOME/bin/Rscript inst/cpp_generation/generate.R
+##$R_HOME/bin/Rscript inst/cpp_generation/generate.R
```
r-ash, line 3: Had to turn this off for now as the structure of some of the generated code has changed. It is going to change again (fairly noisily) I expect as we add the Python interface. So I am tempted to keep this off for now until we settle on the interface again

inst/include/generated/state_types.hpp
```diff
@@ -24,10 +26,32 @@ struct ChildModelState {
...
+template<typename ModelVariant, typename real_type, typename Enable = void>
+struct HivSimulationState {
+  HivSimulationState(const Parameters<ModelVariant, real_type> &pars) {}
+
+  void reset() {}
+};
+
+template<typename ModelVariant, typename real_type>
+struct HivSimulationState<ModelVariant, real_type,
+    std::enable_if_t<ModelVariant::run_hiv_simulation>> {
```
r-ash, line 52: This has to be a little different than the `ChildModel` state as we want this template specialisation to work for `HivFullAgeStratification` and `HivCoarseAgeStratification`. Instead using `run_hiv_simulation` like a trait with `std::enable_if_t`

inst/include/parameters.hpp
```diff
@@ -39,27 +39,44 @@ struct Options {
...
+struct Empty {};
+
 template<typename ModelVariant, typename real_type>
 struct Parameters {
-  BaseModelParameters<real_type> base;
-  ChildModelParameters<ModelVariant, real_type> children;
+    Options<real_type> options;
+    DemographicProjectionParameters<real_type> dp;
+
+    [[no_unique_address]] std::conditional_t<ModelVariant::run_hiv_simulation,
```
r-ash, line 65: `conditional_t<B, T, F>` gives T when B is true, and F when false.

If using C++ 20, with `[[no_unique_address]]` the compiler can optimise the `Empty` struct away to occupy no space. If not C++ 20 then this is just ignored, it should only be a few bytes of allocation which will be pretty negligible compared to what we are doing. See https://en.cppreference.com/w/cpp/language/attributes/no_unique_address

M-Kusumgar: since this is already doing the switch, theres no need for `ChildModelParameters` to be templated on `run_child_model` no? like cant lines 57-64 just be
```cpp
template<typename real_type>
struct ChildModelParameters {
  Children<real_type> children;
}
```

the same being true for `HivSimulationParameters`? they can all share the `Empty` struct

inst/include/generated/state_types.hpp
M-Kusumgar: im not sure i understand this one, does enable if basically add these fields and `reset` function etc if `run_hiv_simulation` is true? how is this different from just defining two structs and using that `conditional_t` to switch between the two at compile time?

M-Kusumgar (commented):
looks pretty good! i had a couple comments but nothing too major!

inst/include/parameters.hpp
r-ash: Yes! Great spot, I'll update that

inst/include/generated/state_types.hpp
r-ash: So if I understand correctly, and I am new to all these concepts to could be missing a point.

Current impl here uses a thing called SFINAE (substitution failure is not an error) which prevents compilation errors when substituting template parameters into a function or class template. So if substitution leads to invalid type of expression, the compiler just ignores that branch. 

The way it works here is the default `HivSimulationState` where `Enable = void` always exists. But if `ModelVariant::run_hiv_simulation` is false then `enable_if_t` causes a substitution failure, so that branch is removed.

We could do this with `conditional_t` but then we'd have to define 2 types. `HivSimulationStateEmpty` and `HivSimulationStateData` for example, and then have the type `HivSimulationState` be a `conditional_t` on the two. So requires a bit more code and means that `HivSimulationStateEmpty` and `HivSimulationStateData` will both be defined in compiled code. With `enable_if_t` if say we are only compiling for the demographic portion of the model (in reality this might never happen!) then only the empty instance of `HivSimulationState` will exist. So does save a bit of compilation time and binary size.

I think ideally we would use new C++ 20 "concepts" see https://en.cppreference.com/w/cpp/language/requires which as I understand work the same as `enable_if_t` but are a bit more explicit about the intention.

Related, I think we could also use this same SFINAE approach for building `HivSimulationParameters` and `ChildModelParameters` but I couldn't work out a non-gross way of instantiating the `Parameters` from these when I tried it this way.

M-Kusumgar (approved):

PR #130, opened 2025-02-07
Remove owned data flag for now

r-ash:
The flag adds a fair bit of complexity, and we're not sure we'll even need it. Indeed we're going to have to copy from C ordered data in Python to Fortran ordered anyway, so we're going to have to copy on the way in and the way out.

This backs it out for now, we can always add it back in if we think we need it.

So I realise in doing this we still need the 2 constructors. 1 from R where we always construct state with 0s. And one from Python. In Python when we want to run year-by-year we want to be passing in the state data. As we need last time step data to have actual values, and indeed if there is some other Avenir module which has run before us (though I don't think this is the case) then we have to use those values.

So actually does beg a question of how we want to handle this, it would be nice to have 1 copy.

At the moment what this is doing is for the State
1. Initialising a big (Fortran ordered) numpy array for each state of 0s
2. Copying a slice of this at time `t` and a slice at time `t - 1` into a `State` object of Eigen::TensorFixedSize on the call to `_initialise_state`. I believe this data is owned by C++ but we should double check it, it will depend on what pybind11 is doing under the hood. (this is the bit of docs I was thinking about when we chatted but I think is only about returning https://pybind11.readthedocs.io/en/stable/advanced/functions.html#return-value-policies)
3. Running for a single time step
4. Copying the data out from the Eigen::TensorFixedSize back into the big numpy array

We should probably chat to Rob G and work out the desired data flow here

### DIFF
--- a/inst/include/child_model_simulation.hpp
+++ b/inst/include/child_model_simulation.hpp
@@ -6,11 +6,11 @@ namespace leapfrog {
 
 namespace internal {
 
-template<typename ModelVariant, typename real_type, bool OwnedData>
+template<typename ModelVariant, typename real_type>
 void run_child_ageing(int t,
                       const Parameters<ModelVariant, real_type> &pars,
-                      const State<ModelVariant, real_type, OwnedData> &state_curr,
-                      State<ModelVariant, real_type, OwnedData> &state_next,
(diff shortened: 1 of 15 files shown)


M-Kusumgar (approved):
LGTM!

Issue #96, opened 2024-08-05
Add note about C++ required version into README


r-ash:
Austin reported some issues installing leapfrog because he had some Makevars set for C++ 11, we should make it clear the dependency on C++ 17. In the README, and in the Makevars for the package.

M-Kusumgar:
we have gone above an beyond and gone to c++20 instead

PR #132, opened 2025-03-07
Mixin struct refactor

M-Kusumgar:
This is the minimum amount of code to get the model working in the new way with the mixins approach. This will serve as an epic branch, base of other improvements to come. This keeps our code more organised, less verbose and helps scientists add new models more efficiently. Main changes include:
* function signature of the rcpp exported function `run_base_model` has changed to better differentiate args that correspond to building the parameters and args that correspond to building the options struct (this may still change in coming PRs)
* moving some 1-based-index logic into the R side because python is 0 index based
* grouping together model functions into structs
* concept based template specialisation to make sure you can only run model code with the correct model variant, e.g. cant run child model simulation without `run_child_model` bool set to `true`
* config struct that outline what parameters they expect + all state and intermediate types they require and mixers that use  conditional inheritance (via template specialisation) to construct the full config of the model
* change to model variant struct to ditch the enum and just have booleans for whether it uses the full stratification or not
* fixed some tests

future PRs:
* code generation using jinja in python using json as our new format instead of csvs
* extracting cpp so that it can be compiled without R or python and the model can be fully run by itself
* add more error handling in cpp (such as checks on max years) and have the cpp function deal with date vector instead of indices
* add ability to input arbitrary initial state and run the model from a particular year
* setting up cmake to copy cpp into R inst include (cran may want the files there) and work with the python interface
* make sure all python interface tests pass
* write docs for how to cpp mixer works and concepts and general project structure
* profit $$$


inst/include/generated/concepts.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+namespace leapfrog {
+
+template<typename ModelVariant>
+concept MV = requires (ModelVariant mv) {
+    { mv.run_demographic_projection } -> std::convertible_to<bool>;
```
plietar: Since these are expected to be static fields, you can use `ModelVariant::run_demographic_projection` instead, and then you don’t even need `(ModelVariant mv)` at all.

R/run_model.R
```diff
@@ -18,9 +18,16 @@ run_model <- function(data, parameters, sim_years,
                       run_hiv_simulation = TRUE,
                       hiv_age_stratification = "full",
                       run_child_model = TRUE) {
+  # TODO Mantra put dates into the cpp instead of having wrappers parse dates
```
r-ash, line 21: Can we make a new issue for this and stick it into the [new project board](https://github.com/orgs/mrc-ide/projects/24/views/1) ? Same for the other TODO below

inst/include/generated/model_variants.hpp
```diff
@@ -1,34 +1,34 @@
...
 
-enum HivAgeStratification {
-    full,
-    coarse,
-    none
-};
+namespace leapfrog {
 
 struct DemographicProjection {
+  static constexpr bool run_demographic_projection = true;
```
r-ash, line 8: Do these have to be bools now? Or would this still support an enum in the future? Just thinking about if there was some 3-way switch we wanted

inst/include/generated/state_space.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+template<MV ModelVariant>
+struct DpSS {
+  static constexpr int NS = 2;
+  static constexpr int pAG = 81;
+};
+
+template<MV ModelVariant>
+struct HaSS {
+  static constexpr std::array<int, 66> hAG_span = {1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
```
r-ash, line 15: C++ is beautiful

src/Makevars
```diff
@@ -1 +1 @@
-PKG_CPPFLAGS = -I../inst/include/
+PKG_CPPFLAGS = -I../inst/include/ -std=c++20
```
r-ash, line 1: Nice! Closes #96 could you add a quick note in the README too?

r-ash (approved):
Mantra this looks amazing, way tidier than what was around before and looks much much simpler to think about and write from the scientists perspective.

~~One thought, not that this needs to be addressed now. If I run with e.g. `run_hiv_simulation = FALSE` and `run_child_model = TRUE` then actually I will just get the demographic projection, now that makes sense why but I wonder if we can return a better message to end users to say why.~~ actually scrap that, we said we'd switch the R interface to specify what model variant to run, instead of passing flags so that avoid this problem

inst/include/generated/concepts.hpp
M-Kusumgar: yep can do

inst/include/generated/model_variants.hpp
M-Kusumgar: theres nothing that explicitly says the flag you put inside the model variant has to be a bool however i was thinking about the code generation and how it would be easier if these were all bools, if there is a legitimate need for a three way switch we would be able to accommodate for that but yh would have to fix up the code generation slightly for that, id be inclined to keep it as bools for now and then if people need this feature, we add it to the code generation then!

inst/include/generated/state_space.hpp
M-Kusumgar: loooool yh love it, so this part is code generated in the next pr, ive keep it hardcoded like this, but we can introduce a utility that helps this in the json later

src/Makevars
M-Kusumgar: oh yep sure

PR #134, opened 2025-03-12
finish code generation for mixin structs

M-Kusumgar:


inst/include/generated/config_mixer.hpp
```diff
@@ -40,101 +40,262 @@ struct ConfigMixer<real_type, ModelVariant> {
...
-  struct Args { \
-    int t; \
-    const Pars& pars; \
-    const State& state_curr; \
-    State& state_next; \
-    Intermediate& intermediate; \
-    const Options& opts; \
-  }; \
-};
+template<typename real_type1, MV ModelVariant1, typename ...Ts>
```
M-Kusumgar, line 44: got rid of the macro, no use for it if the templating language itself has macro capabilities

cpp_generation/src/main.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def generate_hpp(template_name, *args, **kwargs):
+  template = env.get_template(f'{template_name}.j2')
+  output = template.render(*args, **kwargs)
+  with open(relative_file_path(f'../../inst/include/generated/{template_name}.hpp'), "w") as f:
+    f.write(output)
+
+
+dat = load_json("../modelSchemas/FullModel.json")
```
r-ash: For these relative file paths, can we use [importlib_resources](https://importlib-resources.readthedocs.io/en/latest/using.html)?

Perhaps that won't work for the target directory for the code

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -0,0 +1,356 @@
+{
+  "name": "Ha",
+  "namespace": "ha",
+  "add_if": "run_hiv_simulation",
```
r-ash: How about `enable_if`?

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -0,0 +1,356 @@
...
+    {
+      "vars": {
+        "hAG_span": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+        "hDS": 7,
+        "hTS": 3,
+        "hAG": 66
+      }
+    },
+    {
+      "requires": ["use_coarse_stratification"],
```
r-ash: and `applies_if`? Just thinking Rob was a little confused and have to admit this was the bit I found least obvious. Could even make the JSON convey a bit more information

```
"enable_if": "run_hiv_simulation",
"state_space": {
  "default": {
    "vars": ...
  },
  "overrides": [
    {
      "applies_if": "use_coarse_stratification",
      "vars": { ...}
    }
  ]
...
```

What do you think?

r-ash (approved):
This looks great, `inst/include` directory looking nice and tidy! And thanks for chatting this through with me yesterday, all looks great. I agree with your idea to add some schema and validate against this, that'll be super handy. 

Can we add a note into the R package README aimed at Jeff & Maggie about how they should run this. Maybe it just points them to the cpp_generation/README.md

Probably would be nice to just add a script that Maggie & Jeff can use, like the `./scripts/generate` in the R package currently, where we error if `uv` not installed and signpost to the installation docs (or possibly even prompt with a yes/no question and then just install it for them)

We should also remove the code generation from the R package, and the tests. Can do that in another ticket if you want. Could you add an issue for that and put it on the board?

There is also a vignette called `development.Rmd` about how to modify the code, we should update that at some point, could you stick in a ticket for that too?

cpp_generation/src/main.py
M-Kusumgar: we could use this for the json stuff but then if im understanding the docs correctly, we have to put empty `__init__.py` files, that seems less clean, and yh definitely dont want to put that in the generated folder, but hardcoding these paths is bad so maybe ill switch over to `os.path.join` for these paths too?

M-Kusumgar: done

cpp_generation/modelSchemas/configs/HaConfig.json
M-Kusumgar: yep sounds good, done

cpp_generation/modelSchemas/configs/HaConfig.json
M-Kusumgar: yhhh agreed, this was pretty confusing, this is a better structure!

M-Kusumgar:
> This looks great, `inst/include` directory looking nice and tidy! And thanks for chatting this through with me yesterday, all looks great. I agree with your idea to add some schema and validate against this, that'll be super handy.

yep, made a ticket here: https://github.com/mrc-ide/frogger/issues/149

> Can we add a note into the R package README aimed at Jeff & Maggie about how they should run this. Maybe it just points them to the cpp_generation/README.md

have added a small note about this! will expand more once im back on this project

> Probably would be nice to just add a script that Maggie & Jeff can use, like the `./scripts/generate` in the R package currently, where we error if `uv` not installed and signpost to the installation docs (or possibly even prompt with a yes/no question and then just install it for them)

yep thats a good idea, madea ticket for it: https://github.com/mrc-ide/frogger/issues/150

> We should also remove the code generation from the R package, and the tests. Can do that in another ticket if you want. Could you add an issue for that and put it on the board?

done, ticket here: https://github.com/mrc-ide/frogger/issues/151

> There is also a vignette called `development.Rmd` about how to modify the code, we should update that at some point, could you stick in a ticket for that too?

done, https://github.com/mrc-ide/frogger/issues/152

PR #133, opened 2025-03-10
Mkw issue 122 repulled

mwalte10:
This pull request:

**1. Simplified the breastfeeding transmission code**
Breastfeeding previously relied on multiple if/else statements due to how dropout is calculated. Now, dropout is calculated in a separate function, which allows for the breastfeeding function to be more streamlined

**2. Expanded cotrim to work for numerical input when ART is implemented**
Cotrim had been implemented for percentage covered input, and hadn't been tested when ART was also input. This has now been fully tested and aligns

**3. Adds tests (1) to ensure internal consistency between model inputs for the paediatric model and (2) ensure that cotrim need aligns with Spectrum.**
Ensures that population level infections, prevalence, and deaths aligns with the CD4 and transmission/ART timing stratified populations.
Compares cotrim need from Spectrum .dp file with leapfrog's estimate. 

**4. Streamlines the input generation code for testing**
There were previously two separate functions to generate the paediatric input data from the PJNZ file, this has been combined into one function. 

@r-ash and I went through to rebase this so that it can be merged into main, and @M-Kusumgar and I will go through and make sure this doesn't not conflict with the portion he is working on.

PR #147, opened 2025-03-21
Mkw issue 112

M-Kusumgar:
This is maggies branch #133 that I have ported over to the new mixins structs way! Copied PR description from there:

This pull request:

**1. Simplified the breastfeeding transmission code**
Breastfeeding previously relied on multiple if/else statements due to how dropout is calculated. Now, dropout is calculated in a separate function, which allows for the breastfeeding function to be more streamlined

**2. Expanded cotrim to work for numerical input when ART is implemented**
Cotrim had been implemented for percentage covered input, and hadn't been tested when ART was also input. This has now been fully tested and aligns

**3. Adds tests (1) to ensure internal consistency between model inputs for the paediatric model and (2) ensure that cotrim need aligns with Spectrum.**
Ensures that population level infections, prevalence, and deaths aligns with the CD4 and transmission/ART timing stratified populations.
Compares cotrim need from Spectrum .dp file with leapfrog's estimate. 

**4. Streamlines the input generation code for testing**
There were previously two separate functions to generate the paediatric input data from the PJNZ file, this has been combined into one function.


inst/include/models/child_model_simulation.hpp
```diff
@@ -510,110 +511,69 @@ struct ChildModelSimulation<Config> {
...
-    }
+    const auto hPS_dropout_idx = (bf < 6) ? 4 : 5;
+    const auto PMTCT_continuing_rate = 1 - p_hc.PMTCT_dropout(hPS_dropout_idx, t) * 2;
+
+    i_hc.PMTCT_coverage(0) *= PMTCT_continuing_rate; //opt A
+    i_hc.PMTCT_coverage(1) *= PMTCT_continuing_rate; //opt B
+    i_hc.PMTCT_coverage(4) *= PMTCT_continuing_rate; //before pregnancy
+    i_hc.PMTCT_coverage(5) *= PMTCT_continuing_rate; //>4 weeks
+    i_hc.PMTCT_coverage(6) *= PMTCT_continuing_rate; //<4 weeks
+  };
```
mwalte10: Nice, continuing looks weird to me, how about PMTCT_retention ?

M-Kusumgar: sure done!

scripts/create_test_data.R
```diff
@@ -30,8 +30,8 @@ saveRDS(lmod, testthat::test_path("testdata/leapfrog_fit_coarse.rds"))
 mod <- leapfrog::leapfrogR(demp, proj, hiv_steps_per_year = 0L)
 saveRDS(lmod, testthat::test_path("testdata/fit_demography.rds"))
 
-#Create paeds parameters (Run from leapfrog/uncertainrt_analysis_working)
-pjnz_child <- testthat::test_path("testdata/bwa_aim-no-special-elig-numpmtct.PJNZ")
+#Create paeds parameters
+pjnz_child <- testthat::test_path("testdata/bwa_aim-no-special-elig-numpmtct - Copy.PJNZ")
```
r-ash: Why is this file called `-Copy`? Is this the same file as the current file [bwa_aim-no-special-elig-numpmtct.PJNZ](https://github.com/mrc-ide/frogger/blob/main/tests/testthat/testdata/bwa_aim-no-special-elig-numpmtct.PJNZ)?

r-ash (approved):
This looks good, will leave actual model changes for Jeff to review but everything around it looking nice. Looks good to see fewer changed files after the refactor. I wonder if we can have GitHub collapse the generated files in the review? Kind of like it does for large files, that would be neat.

Just one comment on the file names in the test data but pretty minor.

I also added a ticket for putting the "generated by frogger" header back in, not related to these changes #153

scripts/create_test_data.R
M-Kusumgar: i think so, not sure i just checkout maggies pr, ive renamed it so that the copy is removed

mwalte10 (approved):
Thanks for changing the names of the file and the pmtct_retention variable, looks good from my end.

PR #157, opened 2025-04-01
Remove unused code, update test data, ensure benchmark script runs

### LINKED ISSUE #151: remove R code generation + tests
we now have all our code generation in `cpp_generation` so we can remove the old R generation code, remove the tests too

r-ash:
Closes #151 

This PR does a few bits of tidying in the end
1. Removes the R code for code generation
2. Updates the util used in the benchmark script to make sure it still works
3. Update the standalone data generation and run_model script to ensure these work with the new location of metadata

This came from a bit of yak shaving, I almost have the standalone model running again but running into issues with test data related to the metadata location changing. So thought would be good to get this in first and then merge this into my other branch instead of making a huge change


scripts/generate
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-Rscript inst/cpp_generation/generate.R
+cd cpp_generation; uv run src/main.py
```
M-Kusumgar, line 3: maybe just a uv sync too in case
```suggestion
cd cpp_generation; uv sync; uv run src/main.py
```

M-Kusumgar (approved):
LGTM!

scripts/generate
r-ash: I think we can skip this as `uv run` always ensures the project is synced automatically.

https://docs.astral.sh/uv/concepts/projects/sync/#automatic-lock-and-sync

r-ash:
I'll leave this PR open until Jeff has had a chance to look at #147 as otherwise might obfuscate the relevant changes there

scripts/generate
M-Kusumgar: ohhh niceee

PR #158, opened 2025-04-01
Add ability to compile model without C++ headers

### LINKED ISSUE #156: Fix compilation of standalone model
The standalone model exists as a test that we can compile the model without R headers, we turned it off at some point and I think we can no longer do this.

We'll need to do this before we can work on the Delphi interface

After this PR
* I can compile the standalone model without providing R headers
* The tests which run it in R don't necessarily pass, leave this for after the packaging restructure

r-ash:
Closes #156 

This PR will
1. Add support for compiling standalone model (i.e. a model without R headers) and sets a pattern for how we can extend this to support other languages. At the moment I set a preprocessor flag when compiling from R "CALLER_R", you can see this in the `Makevars`, expect will extend these to other flags or use a single flag "CALLER_LANGUAGE" or something which can take values `R`, `cpp`, `delphi`, `python`. For each language we have to define an `InputType` and `OutputType` and functions `parse_data` and `write_output_data`. These are called from the `Config` and so a compile time switch based on what language we are compiling for.

There are a couple of ugly things to think through here, I just wanted to get a version of this up before I went away. I still have to update the code generation to produce this new target, if we're happy with it. One thing here that is really great is how little it has effected the actual model code, credit to @M-Kusumgar on the interface work.

Things I am still unsure about
1. It feels a little weird to me to have the entrypoint `simulate_model` potentially have R specific stuff in it depending on where we compile it from. I feel like it would be a bit cleaner to have the core `simulate_model` function take already prepared `Pars` object instead of creating it via the `Config`. Then we'd be calling the same `simulate_model` from every language. But that would mean we'd probably have some duplication in the interface code, like from leapfrog_py mapping the input names into C++ names, instead of just doing this in the once place. Just feels strange to me to basically have n paths (1 from R, 1 from standalone, 1 from Delphi, 1 from Python), these all come together in call to `simulate_model` and then it have to split out again in calls to `parse_data` and `write_output_data`.


inst/include/frogger.hpp
```diff
@@ -20,26 +22,21 @@ struct Leapfrog {
...
-    const Rcpp::List& data,
+  static void simulate_model(
+    const InputData &data,
     const int time_steps,
     const int hiv_steps,
     const std::vector<int>& save_steps,
     const bool is_midyear_projection,
-    const int t_ART_start
+    const int t_ART_start,
+    OutputData output
```
r-ash, line 32: In the standalone model, `OutputData` is a file path and `build_output` writes to disk. In the model from R `OutputData` is a struct with an `Rcpp::List` and `Rcpp::CharacterVector` of names. To initialize these for standalone model we just want to check that the path exists, and create the directory if not, for R we need to initialize these based on the requested `output_size`. 

So initially I was trying to allocate the `OutputData` in some function within `language_types.hpp`. I was having a bit of a fight (I think) with the list and vector being cleaned up in return from that. Now I'm sure we can work around that, but felt weird to pass in the `output_size` and initialize it like this when the equivalent function for standalone model would be very different.

I just allocated this outside and passed it in to get something working for time being but might revisit this when I get back.

M-Kusumgar (commented):
looks good, I have a couple thoughts about the weird setup and then a potential suggestion:
* I think multiple entrypoints are fine (definitely a good idea as different languages have different interfaces)
* The `get_pars` and building output function live in the config because we are doing the template recursion on them, it feels like a sensible place to put them because its very local to the `Pars` struct definition.
* That being said, i am happy for you to extract them out if it feels weird to have n entry points, then 1 simulate model call and then have n parse pars functions go to 1 run model function but yh then we would need mixers for those two things as well which is fine
* I think grouping these concepts together will be neater, so for example the language specific bits can be in a struct, say `LanguageInterface` for now (naming is up to you XD), so it would have
```C++
#ifdef CALLER_R
struct LanguageInterface {
  struct OutputData {...};
  using InputData = ...;
}
#else
struct LangaugeInterface {
  using OutputData = std::map
  using InputData = std::map
}
#endif
```
(i think `std::map` represents the idea of a general data type better than a file path, i think the serialisation stuff and file paths should be the concern of the cpp file for standalone model rather than the hpp frogger code)

this grouping means you just have to switch out one thing, and also this is a good place to put `get_pars` and `build_model_output` in as well so then you need to make mixers for these functions based on what the model variant is, i.e. `LanguageInterface` would need to be templated on `ModelVariant`. Then you have the big building blocks and `Leapfrog` the entry point class can use this language interface to parse data and build the outputs

This is doing exactly the same thing, maybe with a better level of indirection because you still have n entrypoints, 1 call to `Leapfrog::simulate_model`, n `LanguageInterface`s and 1 `run_model`

I think if you want to go against this pattern of n -> 1 -> n -> 1 calling, youll have to replicate any validation logic/options creation/other checks that may need to happen before parameters are passed but i guess those checks can be extracted into functions!

I was just putting down all my thoughts, you can take or leave as you please!

r-ash:
Closing in favour of #159

Issue #86, opened 2024-07-21
argument names


jeffeaton:
I'm not sure if the `data` and `parameters` arguments to `run_model()` are appropriately descriptively named.

https://github.com/mrc-ide/frogger/blob/29917b3304549ed09346b89b20220b818316e6fa/R/run_model.R#L14-L18

`data` consists of a list of demographic input parameters and `parameters` consists of inputs for the HIV model; see the `demo` and `projp` arguments in Leapfrog: https://github.com/mrc-ide/leapfrog/blob/master/src/leapfrogR.cpp#L7-L8

But there is not really a distinction differentiating one as 'data' and one as 'parameters'.

Also, though, they are just getting concatenated and passed into the model... Maybe not worth separating them out in the input call. (At some point, there was an intention that I could pass only `demp` and it would simulate the demographic projection, without doing any HIV simulation. But we're not really using that and it would be a different 'variant' now)

I suggest we either:
1. Rename these to be more reflective of what they are, or
2. Just combine them into one input list

Lastly, do we have documentation of what needs to be in the lists? If not can we start building that out?

r-ash:
Yes! Let's merge these into one single list.

r-ash:
I want to review this interface more generally, we can merge `data` and `parameters` but what do we want to call the arg, do we also want to merge `sim_years`, `hts_per_year` `output_steps` etc. into some set of options? 

At the moment
* C++ interface is `simulate_model(data, time_steps, hiv_steps, save_steps, is_midyear_projection, t_ART_start)`
* R interface is `run_model<double, ModelVariant>(data, parameters, sim_years, hts_per_year, output_steps, run_hiv_simulation, hiv_age_stratification, run_child_model)`

Should we instead do something in R like
```
run_model <- function(data, options, variant)
```

Where
* `data` is all current `data` and `parameters` combined
* `options` is a list with `sim_years`, `hiv_time_steps`, `output_steps`, `t_ART_start`, `is_midyear_projection`
* `variant` is `DemographicProjection`, `HivFullAgeStratification`, `HivCoarseAgeStratification` or `ChildModel`

For C++ we do something like
```
simulate_model<double, ModelVariant>(data, options)
```

And we try to stick to the same interface for other languages down the line.

r-ash:
Ok perhaps easier to enumerate all the options, makes docs a bit easier to read. So interface then

```
run_model(parameters, configuration, output_years)
```

Questions
[...]

PR #159, opened 2025-04-08
Add ability to compile model without C++ headers

### LINKED ISSUE #156: Fix compilation of standalone model

r-ash:
Closes #156 
This PR adds support for compiling a standalone C++ model. This is another attempt of setting a pattern for this, after trying in #158 and not being super happy with it, this is an alternative approach.

Here the simulation model takes a `Pars` object and returns `OutputState` then the individual interfaces are responsible for converting from a language specific type (e.g. Rcpp::List or python dict) into Pars and are responsible for writing out from `OutputState` into whatever format the interface wants.

For the R and C++ standalone interfaces this defines an "Adapter" (name-wise thinking it adapts the R or C++ data into the data type needed to run the model), these we can generate and use the mixer pattern to combine different requirements for the data

Couple of things I was thinking about but haven't done
1. One downside to this is duplication in the adapter mixing. We could have a single "AdapterMixer" and have the actual adapter used be a preprocessor flag. I didn't go for this though because the actual interface to our adapter functions can be different. We could force them to be the same with some `InputData` `OutputData` indirection like I did in #158 but tbh I think this adds complexity and the code duplication is not too bad. 
2. Could read the data out in the `simulate_model.cpp` interface code, then have the adapter `parse_data` function just create a TensorMap around this read data. This would be nice to mean I don't need the preprocessor flag to switch between tensor types as the simulation model will always work with TensorMaps but the downside is I'd need to know the type of the data I am reading. Which is a little gross, either I'd need just enumerate all the inputs I want to read from disk and their types, or I could loop over the input data dir, read each file to get the type and then build a vector from this. It just felt a bit gross, and worse than just reading it in one place.


r-ash:
Ok addressed some of the things we chatted about, thanks for that. I've updated the adapter mixin to a common one too.

Couple of things which I was thinking about when doing this which would be good to get your input on
1. In the code gen, we could have a common template for the adapter mixers and possibly the adapters themselves. But then we'd lose the matching of locations. At the moment the directory structure of the templates matches that of the generated code, which I think makes it very clear.
2. We had a look yesterday at the interface we need to use when running the model, I think you're right there are some opportunities to improve that and tidy it up. Mostly the magic `0` for the index feels a bit weird to me, but I left changing anything more here for now as we've already made a lot of changes. Think we should do this in another PR, possibly at the same time as looking at #86

inst/include/generated/state_space.hpp
```diff
@@ -5,9 +5,12 @@
 
 #pragma once
 
+#include <array>
```
M-Kusumgar, line 8: why is this necessary?

inst/include/models/child_model_simulation.hpp
```diff
@@ -51,7 +52,7 @@ struct ChildModelSimulation<Config> {
   const State& state_curr;
   State& state_next;
   Intermediate& intermediate;
-  const Options& opts;
+  const Options<real_type>& opts;
```
M-Kusumgar, line 55: definitely optional but the `Options` struct could be in the config like before, cos its just a type but either way is fine!

M-Kusumgar (approved):
Awesome sauce :) got some minor comments but looks good

inst/include/generated/state_space.hpp
r-ash: So I was getting compilation failures when compiling via standalone model. We are using an array here https://github.com/mrc-ide/frogger/pull/159/files/b5f552cccb286715128fd1197ada1174aca22256#diff-455ec435d155f9fab14541e504322200d95630a66448cca0a2e2837adca08504R34 but I wonder if this was being imported by rcpp or something which is why this wasn't an issue when compiling from R? Not sure...

inst/include/models/child_model_simulation.hpp
r-ash: I think I err on the side of keeping it separate if it can be, but I don't feel too strongly if you disagree?

M-Kusumgar: yh im happy either way!

inst/include/generated/state_space.hpp
M-Kusumgar: hmmm yh that would make sense

PR #161, opened 2025-04-09
Issue 156 take2 alternative2

M-Kusumgar:
your perfect forwarding trick and the last attempt involving adapter specialisations on the way out got me thinking what if i could just incorporate this into config mixer itself! it works beautifully!! i think this one is the winner

lets call about this tomo, ill call it a night


r-ash (approved):
Nice one! This looks great, thanks for working this out

Issue #146, opened 2025-03-15
validation for proj years and the time dimensions of the parameters inputted


M-Kusumgar:
currently we just check time dimension of `Sx` in the run_model R function against the length of the sim years but this feels like an arbitrary decision, should there be a source of truth for the max time dimension? what if parameters have different time dimension lengths? do we check all the time dimensions are the same?

r-ash:
Discussed this with Jeff, Maggie and Rob G in weekly meeting. We're happy with this for now. Use `Sx` because it was one of the first things entered into the leapfrog model when Jefff started writing it, and it should always be present. So it is as good a choice as any.

PR #164, opened 2025-04-11
Streamline function interfaces, update usages in tests, and documentation

### LINKED ISSUE #86: argument names

r-ash:
Closes #86 

This PR will
* Update the R interface to `run_model(parameters, configuration, output_years)` and update the C++ interface to match this as closely as possible
* Updated the badges whilst I'm here as we're beyond a concept now, and the build one was showing status from branch builds instead of just main


tests/testthat/test-frogger.R
```diff
@@ -214,63 +196,48 @@ test_that("child model can be run twice on the same data", {
...
+  out <- run_model(input$parameters, "ChildModel", 1970:2030)
+  out2 <- run_model(input$parameters, "ChildModel", 1970:2030)
   expect_identical(out, out2)
 })
 
 test_that("error thrown if trying to run model for more than max years", {
-  demp <- readRDS(test_path("testdata/demographic_projection_object_adult.rds"))
-  parameters <- readRDS(test_path("testdata/projection_parameters_adult.rds"))
+  ## Will error with size of input data not big enough, perhaps this is fine?
+  testthat::skip("TODO: ask science what we want to do here, should we have a limit on this?")
```
r-ash, line 206: @mwalte10 @jeffeaton @M-Kusumgar 

This relates to what we discussed yesterday https://github.com/mrc-ide/frogger/issues/146#event-17220099765

Previously we we're passing in `sim_years` and then if the user entered a value larger than the `year` dimension of `Sx` we were erroring saying "No of years > max years of 61"

Now the `run_model` function is `output_years` and running from `projection_start_time` from the `parameters` object to the largest value in `output_years`. So if they enter `2030` and the projection start time read from the PJNZ is 1970 this will run for 61 time steps. 

We have validation that the input data is the right size, so when you call into C++ it will check `Sx` and every other array has the right size, so if you entered `output_years` as `2040` this would error, unless you also passing in `71` years of data.

Question is: do we want an upper limit on the year still? If so we could hard-code this in the C++ or use `Sx` as the source of truth as we did before.

R/run_model.R
```diff
@@ -1,97 +1,59 @@
...
-    hTS <- dim(data[["art_mort"]])[[1]]
-    data[["h_art_stage_dur"]] <- rep(0.5, hTS - 1)
+  if (is_run_hiv_simulation(configuration)) {
+    hTS <- dim(parameters[["art_mort"]])[[1]]
+    parameters[["h_art_stage_dur"]] <- rep(0.5, hTS - 1)
   }
 
-  run_base_model(data, model_variant, length(sim_years), hts_per_year, output_steps - 1, data[["projection_period"]] == "midyear", data[["t_ART_start"]] - 1)
+  parameters[["is_midyear_projection"]] <-
+    parameters[["projection_period"]] == "midyear"
```
M-Kusumgar: is this logic going to be in python too? if so probably better to keep it in the c++ code, i guess the same thing applies to all the other things we are doing with parameters in this function, perhaps a question for the researchers?

README.md
```diff
@@ -38,9 +35,6 @@ in [`inst/include/frogger.hpp`](inst/include/frogger.hpp). This location
 allows the C++ code to be imported in other R packages via specifying
 `LinkingTo: leapfrog` in the `DESCRIPTION` file.
 
-> [!IMPORTANT]
-> We use C++20 for this package. Please make sure you have a compiler that is compatible.
-
```
M-Kusumgar: do we not want this note anymore?

src/frogger.cpp
```diff
@@ -1,60 +1,90 @@
...
+//' @export
+// [[Rcpp::export]]
+std::vector<std::string> list_model_configurations() {
+  return std::vector<std::string>{
+    "DemographicProjection",
+    "HivFullAgeStratification",
+    "HivCoarseAgeStratification",
+    "ChildModel"
+  };
+}
```
M-Kusumgar, line 22: do we expect different languages to have different values for `list_model_configurations`? if so, this is fine but i think probably worth, in another ticket maybe, code generating all the available configurations (as these are derived from the json files) and this can be a general c++ function and the entrypoints to the language specific cpp files reference that function

src/frogger.cpp
```diff
@@ -1,60 +1,90 @@
...
 
-  const auto opts = leapfrog::get_opts<real_type>(hiv_steps, t_ART_start, is_midyear_projection);
-  const auto pars = LF::Cfg::get_pars(data, opts, time_steps);
+  const int t_art_start = Rcpp::as<int>(parameters["t_ART_start"]);
+  const bool is_midyear_projection = Rcpp::as<bool>(
+    parameters["is_midyear_projection"]);
+  int hts_per_year = 10;
+  if (parameters.containsElementNamed("hts_per_year")) {
+    hts_per_year = Rcpp::as<int>(parameters["hts_per_year"]);
+  }
```
M-Kusumgar: if we are in the R specific entrypoint anyway, i feel like null/undefined/not found logic should be in R itself as it is an easier language to write, in general i think null/undefined/not found logic should probably go in the wrapper for the cpp function in the languages themselves like `run_model` in `run_model.R` file

tests/testthat/test-frogger.R
M-Kusumgar: from a coding perspective i dont see a reason to artificially put an upper limit on the year if people have the data, in 20 years time people can still run predictions if they want

M-Kusumgar (commented):
Looks good on the whole! Had a couple of small comments

R/RcppExports.R
```diff
@@ -1,8 +1,16 @@
 # Generated by using Rcpp::compileAttributes() -> do not edit by hand
 # Generator token: 10BE3573-1514-4C36-9D1C-5A225CD40393
 
-run_base_model <- function(data, model_variant, time_steps, hiv_steps, save_steps, is_midyear_projection, t_ART_start) {
-    .Call(`_frogger_run_base_model`, data, model_variant, time_steps, hiv_steps, save_steps, is_midyear_projection, t_ART_start)
+#' List the avaialble model configurations
```
M-Kusumgar, line 4:
```suggestion
#' List the available model configurations
```

R/run_model.R
```diff
@@ -1,97 +1,59 @@
...
-#' @param data Input data
 #' @param parameters Projection parameters
-#' @param sim_years Simulation years to run model for default 1970:2030
-#' @param hts_per_year Number of HIV time steps per year, default 10
-#' @param output_steps Which sim years to output for default same as sim_years
-#' @param run_hiv_simulation If TRUE then runs HIV simulation
-#' @param hiv_age_stratification The age stratification for HIV population,
-#'  "coarse" or "full"
-#' @param run_child_model If TRUE then run the child model
+#' @param configuation The model configuration to run, see
```
M-Kusumgar:
```suggestion
#' @param configuration The model configuration to run, see
```

README.md
r-ash: Oh nope! We do indeed want it, got wiped by generating this from the `Rmd`. I've added it back in

src/frogger.cpp
r-ash: Yeah so I did think about this, I suppose it can be different depending on what it was compiled with. I expect in reality Python will have all of them, R will have all of them and Delphi will have just 1 but TBC

src/frogger.cpp
r-ash: That's a good shout, I'll move this over

R/run_model.R
r-ash: Thanks, yes think it will be, I've tidied that up and moved it into common C++

r-ash: re other bits, probably some of them.

```
hTS <- dim(parameters[["art_mort"]])[[1]]
parameters[["h_art_stage_dur"]] <- rep(0.5, hTS - 1)
```

This looks generic to me, or if not we should do it in the setup. Let's see what @mwalte10 and @jeffeaton think.

The other things we have here are
1. Converting indexes from 1-based to 0-based is R-specific
2. Converting from names `_coarse` and `_full` we might want this in Python. It depends where that data comes from I think. @mwalte10 or @jeffeaton is both the `full` and the `coarse` always in the PJNZ or only 1 ever present and you just create it as `_full` or `_coarse`? Is it worth collapsing this to just use a single name to remove this renaming weirdness?

tests/testthat/test-frogger.R
jeffeaton: I'm not 100% following the question here. 

> We have validation that the input data is the right size, so when you call into C++ it will check Sx and every other array has the right size, so if you entered output_years as 2040 this would error, unless you also passing in 71 years of data.

This sounds like correct / intended behaviour to me.

> Question is: do we want an upper limit on the year still? If so we could hard-code this in the C++ or use Sx as the source of truth as we did before.

I wasn't aware we had a hard coded upper limit on the years. This always came from how many years of demographic and HIV programme inputs you provided. Which I think is the same as no hard coded upper limit, but upper limit comes from `Sx` (but that is my understanding of the behaviour described above, so concerned I'm missing some nuance or missing the point).

R/run_model.R
jeffeaton: PJNZ always represents `_full`.  We create `_coarse` inputs by collapsing PJNZ inputs accordingly.

Where are you want to to collapse? (I'm not seeing a `_full` and `_coarse` represented in the highlighted code.)

tests/testthat/test-frogger.R
r-ash: This is great, sorry for the confusion that is my fault for misreading/misremembering. There wasn't an upper limit, you're right, I was getting confused with some special handling I had put into the standalone model but that had no affect on what was here in the R interface. Actual difference here is that the validation we now have is that dimensions of all input data is correct, nothing special about years. This all sounds good then

R/run_model.R
r-ash: Can see here https://github.com/mrc-ide/frogger/blob/main/R/run_model.R#L63 

So trying to decide if this is the way we always want to pass data in (from all interfaces) and so should we move it into common C++ code. Or is it something specific about R and we want to keep, or specific about R but we can remove it.

r-ash: Ok chatted this through we Jeff, we want to keep the `_full` and the `_coarse` switching and setting `h_art_stage_dur` and we want them for all interfaces. So let's move them into the C++

src/frogger.cpp
M-Kusumgar: hmmm okay, for now happy to leave this in here but perhaps we should think of a less duplicated approach later, its quite brittle here since all it takes is someone changing the name of the model variant to break this whole interface, i think it should somehow be linked to the code generation but for another PR in the future!

M-Kusumgar (approved):

R/run_model.R
r-ash: Added an issue to address this later #168

PR #166, opened 2025-04-11
Add script for prompting user to install uv when they run generate

### LINKED ISSUE #150: add script for code generation
it would be nice to add a script that jeff or maggie can run `./scripts/generate` that will check if uv is installed (if not give them a yes/no prompt and if yes install it for them maybe) and sync the python packages and run the code generation

M-Kusumgar:
Closes #150 
This just updates the script so it checks if a user has `uv` installed and if not then it prompts them and tries `wget` if that fails it tries to look for `pip` and lastly if that fails, it gives the user a link to the installation docs

### DIFF
--- a/scripts/generate
+++ b/scripts/generate
@@ -1,12 +1,21 @@
 #!/usr/bin/env bash
 
 if ! uv --version >> /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


r-ash (approved):
This looks great, I updated the "Description" with "Closes #150" which will close the issue for us automatically

PR #167, opened 2025-04-15
Add support for initial state input

### LINKED ISSUE #144: add ability to input initial state into the model to allow running from arbitrary state
allow for input initial state remembering to disable p_total_pop in run_model function

M-Kusumgar:
Closes #144 

I have taken a similar approach to Pars parsing from different languages and applied to state. I have tried to reduce code duplication as much as possible but still a bit unsure about the interface! Let me know what you think.

### DIFF
--- a/R/run_model.R
+++ b/R/run_model.R
@@ -7,12 +7,18 @@
 #'   defaults to all years from 1970 to 2030. Also used to control what years
 #'   the simulation is run for. If output only 2030, simulation will be run
 #'   from `projection_start_year` passed in the `parameters` list.
+#' @param initial_state If provided, the model will run from this initial state
+#'   usually `start_from_year` should also be specified with this. (default NULL)
+#' @param start_from_year If provided, start model simulation from a particular
+#'   this year. (default 1970)
 #'
 #' @return List of model outputs
 #' @export
 run_model <- function(parameters,
                       configuration = "HivFullAgeStratification",
-                      output_years = seq(1970, 2030)) {
+                      output_years = seq(1970, 2030),
+                      initial_state = NULL,
+                      start_from_year = 1970) {
 
   if (configuration == "HivCoarseAgeStratification") {
     parameters$hAG_SPAN <- parameters[["hAG_SPAN_coarse"]]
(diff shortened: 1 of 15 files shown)


R/run_model.R
```diff
@@ -49,11 +55,39 @@ run_model <- function(parameters,
...
+
+concat_on_time_dim <- function(dat1, dat2) {
+  x <- lapply(names(dat1), function(name) {
+    con <- abind::abind(dat1[[name]], dat2[[name]])
+    dimnames(con) <- NULL
+    con
+  })
+  names(x) <- names(dat1)
+  x
+}
```
r-ash, line 113: Can we move these into test helpers? Looks like they're only used in testing atm?

src/frogger.cpp
```diff
@@ -21,12 +21,14 @@ std::vector<std::string> list_model_configurations() {
...
 }
 
-template<typename real_type, typename ModelVariant>
+template<typename ModelVariant>
 Rcpp::List simulate_model(
   const Rcpp::List parameters,
-  const std::vector<int> output_years
+  const std::vector<int> output_years,
+  const std::optional<Rcpp::List> initial_state_data,
+  std::optional<int> start_from_year
```
r-ash: Do we not want this templated on `real_type` for TMB impl? Do you know what entrypoint TMB actually uses?

R/run_model.R
```diff
@@ -7,12 +7,18 @@
...
 #'   the simulation is run for. If output only 2030, simulation will be run
 #'   from `projection_start_year` passed in the `parameters` list.
+#' @param initial_state If provided, the model will run from this initial state
+#'   usually `start_from_year` should also be specified with this. (default NULL)
+#' @param start_from_year If provided, start model simulation from a particular
+#'   this year. (default 1970)
 #'
 #' @return List of model outputs
 #' @export
 run_model <- function(parameters,
```
r-ash, line 13: I prefer a separate function here instead of using the existing one. I think it makes what the functions do easier to understand and also advertises better the fact that you can run it from a particular state. Happy to discuss if you feel different. I also like that it makes the docs simpler.

We could have
```
run_model_from_state(parameters, configuration, state, start_year, output_years)
```

And then down the line
```
run_model_single_year(parameters, configuration, state, start_year)
```

inst/include/frogger.hpp
```diff
@@ -23,25 +23,34 @@ struct Leapfrog {
   static OutputState run_model(
     const Pars& pars,
     const Options<real_type>& opts,
+    const std::optional<State> initial_state,
```
r-ash: I think I'd also be tempted to try this as 2 separate functions

```
run_model_from_state(pars, opts, state, start_year, output_years)
```

And have these 2 be the exported functions. `run_model` could create the initial state and then call `run_model_from_state`

tests/testthat/test-child-model.R
```diff
@@ -358,3 +358,17 @@ test_that("HIV related deaths among CLHIV on ART align", {
 
   expect_true(all(abs(dt$diff) < 5e-2))
 })
+
+test_that("Child model agrees when run in single years vs all years", {
```
r-ash, line 362: Nice!

r-ash (changes requested):
This looks great, I wonder about using 2 functions instead of adding the additional args to `run_model`? What do you think?

R/run_model.R
M-Kusumgar: i dont mind moving into test helpers but i was thinking these are useful for anyone running the model in this way, like both these functions are necessary for continuing the function from a particular state and aggregating the result, not sure they are just test helpers

src/frogger.cpp
M-Kusumgar: so i thought about TMB and since TMB doesnt follow the same parsing that we follow in our R adapter, which means it needs its own adapters so we wont be using this entrypoint for it

R/run_model.R
M-Kusumgar: i dont mind the separate function here, i was half on the edge of this, but no strong opinions! i had it separate at first but they share most of the code, ill just pull that out into another function as well

inst/include/frogger.hpp
M-Kusumgar: after thinking about this for a while, i think i agree with you, will make this change

R/run_model.R
r-ash: That makes sense. Maybe we should wrap the output result in a s3 type and then have these check that to formalize the interface a bit, and document and export these. We can do that down the line.

r-ash: Added an issue for that #169

r-ash (approved):
This looks great

PR #170, opened 2025-04-16
Clean up common input parsing

### LINKED ISSUE #168: Move common input parsing into C++
In the R `run_model` function we have some input parsing. Some of this is specific to R e.g. converting from 1-based counting to 0-based, others is not

See https://github.com/mrc-ide/frogger/pull/164#discussion_r2044111520 for a discussion

TODO:
1. Move over renaming of `_full` or `_coarse` variables into common C++ code we can use from every lanaguage
2. Move setting of `h_art_stage_dur` into C++ code so we can use everywhere, perhaps error if this is also passed in by the user? As it's hardcoded so always going to be overwritten

M-Kusumgar:
Closes #168 

Two main problems to tackle in this PR were:
1. `_coarse` and `_full` suffices for parameters like `art_mort`
2. `h_art_stage_dur` being constant

For the first problem I have generalised our overrides and default json struct to the parameters config and so now we support template specialisation on parameters based on model variant booleans. There is a slight change in structure and behaviour, there is no `"default": { "vars": {...} }` structure, instead just `"default": {...}` as this is terser, and now you only have to specify the vars you actually want to override, it will fill in all default vars automatically for easier override definitions and less duplication, The python function ive written also means it should be pretty easy to support default and override in other sections too if we need

For the second problem it doesnt make sense to me to have a parameter input that is always constant, if it will change in the future then we make it a full parameter and for the time being set it as `0.5` in the R function or if it wont change in the future we set it as a state space variable.


inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -37,6 +37,7 @@ struct AdultHivModelSimulation<Config> {
   static constexpr int MALE = SS::MALE;
   static constexpr int FEMALE = SS::FEMALE;
   static constexpr int ART0MOS = SS::ART0MOS;
+  static constexpr auto h_art_stage_dur = SS::h_art_stage_dur;
```
M-Kusumgar: @jeffeaton @mwalte10 it doesnt make sense to me to have a parameter input that is always constant, if it will change in the future then we make it a full parameter and for the time being set it as 0.5 in the R function or if it wont change in the future we set it as a state space variable, what do you think?

for context this was being set as `rep(0.5, hTS - 1)` in the R code

r-ash (approved):
Nice this looks great to me, very clean! I'll leave this open for @jeffeaton to have a quick check, but all good from my end

inst/include/models/adult_hiv_model_simulation.hpp
r-ash: Let's move `h_art_stage_dur` into an input parameter again, for the others

But we can move these into compile time, these will be the same for every currently known about model variant

p_idx_fertility_first(15),
p_fertility_age_groups(35),
p_idx_hiv_first_adult(15),
adult_incidence_first_age_group(15),
pAG_INCIDPOP(35),
hIDX_15PLUS(0),

Just noting for the future: this came from Jeff request. `h_art_stage_dur` "feels like" a parameter, so possible we might want to be able to change it in the future at runtime.

r-ash (approved):
This looks excellent!

PR #172, opened 2025-04-23
Add prototype of C dll interface we can call from Delphi

r-ash:
This PR adds a prototype of how we can call leapfrog from Delphi. It only exposes the demographic projection model for the time being. I want to make a plan from this and ticket up the remaining work. Couple of things worth chatting about

1. This exposes a struct to the Delphi code, I do like this and it helps in delphi so that we don't need to pass in a huge array of args to the actual function. I can see this going 2 ways, expose struct with items specific to what spectrum desktop wants to pass i.e. the type of model it wants to run. Or expose our existing param structs which exist on the config, if this is possible.
2. There are a few hardcoded things here, `output_years`, `projection_period`, `t_ART_start`, `hts_per_year` which should come from the calling code
3. The `build_dll.sh` script copies the DLL into the location I have spectrum repo at relative to this repo, not super portable and we should think of something better for that


r-ash:
Now implemented for real so closing this prototype

PR #176, opened 2025-04-23
Add support for single year runs

### LINKED ISSUE #173: Add a function to run a single year of the model from some state input
And output for a single year.

We'll want to call this from Delphi and Python.

Do we also need to expose a set initial state function? At the moment all this does is populate the first year state from the base population, but it might do more in the future? Could just put that in another ticket and we can write it when we need it.

M-Kusumgar:
Closes #173 

This adds support for single year runs. Was a bit fiddly, I added support to only the R interface because i dont know if we want to use it in cpp, i think the cpp is more like a smoke test interface

I ended up using function overloading the the `frogger.cpp` file, i think this is cleaner and got rid of the last remaining bits of `std::optional`s


r-ash (approved):
This looks good! I see the fiddlyness in `frogger.cpp` but this looks like a good approach. Let's double check the interface in the meeting today

R/run_model.R
```diff
@@ -42,6 +42,25 @@ run_model_from_state <- function(parameters,
...
 
+#' Run leapfrog model fit for a single year
+#'
+#' @param parameters Projection parameters
+#' @param configuration The model configuration to run, see
+#'   [list_model_configurations()] for available configurations
+#' @param initial_state The model will run from this initial state
+#' @param start_from_year Start the model simulation from this year
+#'
+#' @return List of model outputs
```
r-ash: We should make clearer how the return type from this is different to above. And when would I want to run this vs the other run model types.

We could document all these together too.

M-Kusumgar: think ive addressed this

PR #182, opened 2025-04-25
Fix small bug in input generation, and regen all input parameters

r-ash:
I think this fixes a small issue in input parameter when creating `births_sex_prop`

We are taking `demp$srb` which is a 1D array of length 61, values from 1970 to 2030.

We are building `births_sex_prop` by doing

```
rbind(male = demp$srb, female = 100) / (demp$srb + 100)
```

I believe what this is doing is dividing male 1970 by `demp$srb[1] + 100`, female 1970 by `demp$srb[2] + 100`, male 1971 by `demp$srb[3]` and so on then reusing the vector. Instead of what I think we intend which is diving male 1970 by `demp$srb[1] + 100`, female 1970 also by `demp$srb[1] +100` and so on.

Here is an example of what I mean with some dummy data

```
Browse[1]> my_data <- data.frame(a = c(2, 3), b = c(4, 5))
Browse[1]> my_data
  a b
1 2 4
2 3 5
Browse[1]> my_divisor <- c(6, 7)
Browse[1]> my_data / my_divisor
          a         b
1 0.3333333 0.6666667
2 0.4285714 0.7142857
Browse[1]> data.frame(a = c(2 / 6, 3 / 7), b = c(4 / 6, 5 / 7)) // note the denominator
          a         b
1 0.3333333 0.6666667
2 0.4285714 0.7142857
```

I noticed this when looking at how spectrum desktop was preparing this input. This uses an approach like spectrum desktop.


jeffeaton:
Nice catch. I guess that never failed on a test because the SRB was constant in all our test files.

Thanks

jeffeaton (approved):

PR #185, opened 2025-05-13
Add Delphi interface

### LINKED ISSUE #181: Finalise structure for the Delphi interface and generate it
See https://github.com/mrc-ide/frogger/pull/172
And https://github.com/AvenirHealth-org/Spec5/pull/6 for the prototype

How do we want to extend this for rest of the parameters and inputs that leapfrog needs?

I think couple of options:
1. Continue with the named record that we have and translating that into the DLL
2. Pass an arbitrary named list of inputs

The first will require a bit more book keeping and a bit more boiler plate code, but it might be less error prone as we'll be advertising to Delphi exactly what inputs need to be provided. Second will be more flexible but easier to get stuff wrong.

Let's also think about how we can verify sizes of arrays, it would be nice to be able to check this (perhaps only when in development mode?) and error if the passed in data is the wrong size. Few ways we could do it:
1. Generate the dimensions into the Delphi wrapper, and when we assign the array make sure it is the correct length (we won't be able to check the runtime dimension, the number of projection years)
2. Pass the dims from Delphi into C++ and have C++ assert the size is right
3. Detect when the end of the array is? (using some null terminated array?) and then check full size is what we expect

r-ash:
Closes #181 
Sorry this is a bit chunky, I can break it up if it is a bit too much.

This PR will
* Add a generated "c" interface. This is an interface with c-linkage specification. In practice we'll call this from Delphi and the interface is specific to Delphi, but it could be called from other languages. 
   * There are 4 ways to run the model included here, every year and single year for demographic projection only and every year and single year for child model. I've only tested the every year demographic projection. I think before we can use single year we'll need a "initial year events" function which will take params and return state
* Add the Delphi `.pas` generated code counter part. Also included here is the tensor fixed size helper which will need to exist for the Delphi impl to work. It is not generated, and I don't expect it to change much but included for completeness

I am expecting to add a script into the delphi code which will
1. Clone this repo
2. Build the dll (see the `./c_interface/build_dll.sh` script)
3. Copy the dll into the required location
4. Copy the generated delphi code into the correct location

We can have this script use a branch pin, and use this to allow spectrum desktop to decide when it pulls in changes to leapfrog.

Something that would be nice to add here would be a test, now I think this will be difficult to do of the `pas` code, indeed I don't think there is a way to run Delphi on CI. But we could probably test the c interface, or maybe there is a way to run the `.pas` code with open pascal. Something to look into.

Next steps
1. Add a PR to spectrum desktop to clone this down and to call leapfrog DLL
2. Chat with Rob about the options we want exposed to Delphi, at the moment it is only start and end year
3. Look into testing the interface, at a minimum add a CI build which builds the DLL
4. Expose a "initial year" events function to Delphi


c_interface/c_interface.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+#define DllExport extern "C"
+#define EXPORT comment(linker, "/EXPORT:" __FUNCTION__ "=" __FUNCDNAME__)
+
+typedef void (WINAPI *CallbackFunction)(const char*);
+
+template <typename ModelVariant>
+HRESULT fit_model(leapfrog::internal::CParams<double> &data,
+                  leapfrog::internal::COptions &options,
+                  leapfrog::internal::COutput<double> &out,
```
r-ash: Lets rename this `CState` instead.

Also because we are just passing pointers about, this can be output state or single state. I expect down the line we won't need the `OutputState` i.e. the state with an extra dimension for time. If we do want both, might be sensible to have separate types?

c_interface/c_interface.cpp
```diff
@@ -0,0 +1,118 @@
...
+    std::vector<int> output_years((options.proj_end_year - options.proj_start_year) + 1);
+    std::iota(output_years.begin(), output_years.end(), options.proj_start_year);
+
+    const leapfrog::Options<double> opts =  {
+      10,
+      34,
+      leapfrog::internal::BaseSS::PROJPERIOD_CALENDAR,
+      options.proj_start_year,
+      options.proj_end_year
+    };
```
r-ash: Need to chat to Rob to work out what options we want exposing to Delphi

c_interface/c_interface.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    auto state = LF::run_model(pars, opts, output_years);
+    LF::Cfg::build_output(0, state, out);
+  } catch (const std::invalid_argument& e) {
+    error_handler(e.what());
+    return E_INVALIDARG;
+  } catch (...) {
+    error_handler("Caught unhandled exception");
+    return E_FAIL;
+  }
```
r-ash, line 44: We should expand this for other potential error types, need to review specific errors we throw

cpp_generation/templates/delphi/LeapfrogInterface.j2
```diff
@@ -0,0 +1,113 @@
...
+type
+  Leapfrog{{ config.name }}Params = record
+  private
+    {% for name, cfg in config.pars.default.items() %}
+    {{ to_lower_camel_case(name) }}: {{ get_delphi_ptr_type(cfg) }};
+    {{ to_lower_camel_case(name) }}Length: Integer;
+    {% endfor %}
+  public
+    {% for name, cfg in config.pars.default.items() %}
+    procedure Set{{ to_camel_case(name) }}(var in{{ to_camel_case(name) }}: TGBFixedArray<{{ get_delphi_num_type(cfg) }}>);
```
r-ash: Quite hard to read, perhaps a helper function for this?

c_interface/c_interface.cpp
r-ash: Added an issue #186

c_interface/c_interface.cpp
r-ash: #187

c_interface/c_interface.cpp
r-ash: Done!

cpp_generation/templates/delphi/LeapfrogInterface.j2
r-ash: Tidied this up

delphi/LeapfrogInterface.pas
```diff
@@ -0,0 +1,955 @@
...
+    procedure SetCtxNeed(var inCtxNeed: TGBFixedArray<Double>);
+    procedure SetInfectionByType(var inInfectionByType: TGBFixedArray<Double>);
+end;
+
+{$ALIGN 8}
+type
+  LeapfrogParams = record
+  private
+    dp: ^LeapfrogDpParams;
+    ha: ^LeapfrogHaParams;
```
r-ash: We have a `ha` module in Spectrum already so this might be a bit confusing. Can we use long names for these? "HivAdult", "HivChild", "DemProj"

delphi/LeapfrogInterface.pas
```diff
@@ -0,0 +1,955 @@
...
+    procedure SetHaState(var haState: LeapfrogHaState);
+    procedure SetHcState(var hcState: LeapfrogHcState);
+end;
+
+type TCallbackFunction = procedure(Msg: PAnsiChar); stdcall;
+
+procedure LeapfrogRunDp(var params: LeapfrogParams; var opts: LeapfrogOptions; var outputState: LeapfrogState; errorHandler: TCallbackFunction); safecall; external 'leapfrog.dll' name 'run_dp';
+procedure LeapfrogRunAim(var params: LeapfrogParams; var opts: LeapfrogOptions; var outputState: LeapfrogState; errorHandler: TCallbackFunction); safecall; external 'leapfrog.dll' name 'run_aim';
+procedure LeapfrogRunDpSingleYear(var params: LeapfrogParams; var opts: LeapfrogOptions; var initial_state: LeapfrogState; start_year: Integer; var state: LeapfrogState; errorHandler: TCallbackFunction); safecall; external 'leapfrog.dll' name 'run_dp_single_year';
+procedure LeapfrogRunAimSingleYear(var params: LeapfrogParams; var opts: LeapfrogOptions; var initial_state: LeapfrogState; start_year: Integer; var state: LeapfrogState; errorHandler: TCallbackFunction); safecall; external 'leapfrog.dll' name 'run_aim_single_year';
```
r-ash: We need a function to set up this `initial_state` for the first year. See what leapfrog full model does, this is separate from the base year calculation in spectrum desktop to back project in the preceding year.

c_interface/c_interface.cpp
r-ash: Actually, done this here!

delphi/LeapfrogInterface.pas
r-ash: This was from Rob G, I've made this change now

delphi/LeapfrogInterface.pas
r-ash: Actually not 100% if we do need this, something to chat through. I think if we have the case that
1. We do some spectrum desktop "stuff"
2. We project the first year of leapfrog
3. We do additional spectrum desktop "stuff"
4. We project 2nd year of leapfrog 
5. etc.

We'll want a way to create the state in Spectrum desktop, because they will do some intermediate work, and we'll need to map that data into the state. I doubt it will work on the same data as the state sadly (at least for now).
But then the question is whether we should expose a helper to set up the state the first time, because when we run the first year all we need is the total pop from last year, everything else should be 0. I suppose this might change in the future, so perhaps we should expose it. Worth chatting through this but I'll expose this function for now.

r-ash: Added an issue for this #189 I am thinking that probably this will all get handled in spectrum desktop down the line once we work out how to run year-by-year. But let's discuss it.

cpp_generation/modelSchemas/configs/DpConfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "Dp",
+  "delphi_name": "DemProj",
```
M-Kusumgar: i think delphi names should follow a pattern and not be an extra field in the metadata, only because if a researcher wants to add another config and they only care about the R interface, they shouldnt have to think about the delphi name

can probably just redefine whatever is generated out to any name you want in the delphi code

c_interface/c_interface.cpp
```diff
@@ -0,0 +1,118 @@
...
+                              leapfrog::internal::CState<double> &out,
+                              CallbackFunction error_handler) {
+
+  using LF = leapfrog::Leapfrog<leapfrog::C, double, ModelVariant>;
+
+  try {
+    const leapfrog::Options<double> opts =  {
+      10,
+      34,
+      leapfrog::internal::BaseSS::PROJPERIOD_CALENDAR,
```
M-Kusumgar: pass in options here

cpp_generation/modelSchemas/configs/DpConfig.json
M-Kusumgar: can just change the field to long name

M-Kusumgar (approved):

PR #197, opened 2025-05-16
Unskip reference tests that check output matches leapfrog

### LINKED ISSUE #97: Remove the skip on reference tests
After #94 we should be able to update the test data and the reference tests should work

r-ash:
Closes #97 

Since #94 these should match, this gets around to unskipping those tests, and fixes a small typo in reference generation which was causing this to fail.

There are a couple of remaining skips, I wonder if we can remove them.

@mwalte10 There is a skip on one of the child tests https://github.com/mrc-ide/frogger/blob/679e80b4e2c3b20443b8ebfdcd54355448ee4ff4/tests/testthat/test-child-model.R#L115 should that be matching now since #94. I double checked and it is still failing but I wonder if that is expected or should it be passing now?

@M-Kusumgar There is a skip here too https://github.com/mrc-ide/frogger/blob/679e80b4e2c3b20443b8ebfdcd54355448ee4ff4/tests/testthat/test-frogger.R#L206 I feel like we did talk about that with Jeff, Maggie and Rob do you recall the outcome of that? I think it is fine to be able to project onward right?


mwalte10:
I've just checked and that test compares against the Spectrum output rather than the stored estimates from the old leapfrog repository. I've just checked and it looks like the female HIV population 15-49 isn't aligned with the Spectrum software output.  We should raise this with Jeff, sorry would have earlier but forgot this test was on!  You can recreate this misalignment using the following code: 

pjnz_adult <- system.file("pjnz/bwa_aim-adult-art-no-special-elig_v6.13_2022-04-18.PJNZ", package = "frogger", mustWork = TRUE)
dp = eppasm::read_dp(pjnz_adult)
parameters <- readRDS(test_path("testdata/adult_parms.rds"))
out <- run_model(parameters)

spec <- SpectrumUtils::dp.output.hivpop(dp, direction = 'long')
spec <- spec %>%
  dplyr::filter(Age %in% 15:49 & Sex == 'Female') %>%
  dplyr::group_by(Year)

lfrog <- out$p_hiv_pop[16:50,2,]
dimnames(lfrog) <- list(Age = 15:49, Year = 1970:2030)
lfrog <- as.data.frame(as.table(lfrog)) %>%
  dplyr::mutate(lfrog = Freq)
lfrog$Year <- as.integer(as.character(lfrog$Year))


dt <- dplyr::right_join(spec, lfrog, by = c("Year", "Age"))
dt <- dt %>%
  dplyr::mutate(diff = Value - lfrog)

r-ash:
> I've just checked and that test compares against the Spectrum output rather than the stored estimates from the old leapfrog repository. I've just checked and it looks like the female HIV population 15-49 isn't aligned with the Spectrum software output. We should raise this with Jeff, sorry would have earlier but forgot this test was on!

Thanks Maggie, I'll add this to agenda for next catch up

tests/testthat/test-reference.R
```diff
@@ -11,15 +9,6 @@ test_that("demographic model is correct", {
...
   expect_equal(actual$p_total_pop_background_deaths, expected$natdeaths)
-  expect_equal(actual$p_hiv_pop, expected$hivpop1)
-  expect_equal(actual$p_hiv_pop_background_deaths, expected$natdeaths_hivpop)
-  expect_equal(actual$h_hiv_adult, expected$hivstrat_adult)
-  expect_equal(actual$h_art_adult, expected$artstrat_adult)
-  expect_equal(actual$h_hiv_deaths_no_art, expected$aidsdeaths_noart)
-  expect_equal(actual$p_infections, expected$infections)
-  expect_equal(actual$h_hiv_deaths_art, expected$aidsdeaths_art)
-  expect_equal(actual$h_art_initiation, expected$artinit)
-  expect_equal(actual$p_hiv_deaths, expected$hivdeaths)
```
M-Kusumgar, line 22: im confused, why are we not checking these?

M-Kusumgar (commented):
> I feel like we did talk about that with Jeff, Maggie and Rob do you recall the outcome of that? I think it is fine to be able to project onward right?

I thought the consensus was to not have a max number of years, thats what i recall at least

tests/testthat/test-reference.R
r-ash: We don't generate these any more, since a nice refactor happened :eyes:, they we're previously all 0

r-ash:
> > I've just checked and that test compares against the Spectrum output rather than the stored estimates from the old leapfrog repository. I've just checked and it looks like the female HIV population 15-49 isn't aligned with the Spectrum software output. We should raise this with Jeff, sorry would have earlier but forgot this test was on!
> 
> Thanks Maggie, I'll add this to agenda for next catch up

We chatted about this, this is a known difference between the models that Jeff is working on. We should be able to remove it soon!

PR #201, opened 2025-05-23
Tidy up generation code, ensure scalars are passed as such from Delphi

r-ash:
In Spectrum desktop I am extending the interface to call the adult HIV model. There was a bit of difficulty with passing scalars as the codegen was expecting them to be passed as pointers. I've tidied this up so they can just be passed as scalars which simplifies the calling code. I also split up the code gen helpers a bit.

This does not close 195 but is required for it


c_interface/c_interface.cpp
```diff
@@ -59,7 +59,7 @@ DllExport HRESULT WINAPI run_aim(leapfrog::internal::CParams<double> &data,
    leapfrog::internal::CState<double> &out,
    CallbackFunction error_handler) {
 #pragma EXPORT
-return fit_model<leapfrog::ChildModel>(data, options, out, error_handler);
+return fit_model<leapfrog::HivFullAgeStratification>(data, options, out, error_handler);
```
r-ash, line 62: Putting this here temporarily but will be undone down the line

M-Kusumgar (approved):
Looks good to me! could you put the generated files in the delphi folder in the gitattributes so that they are disabled for the diff too? (it was helpful this time but best to disable github from diffing those by default)

r-ash:
> Looks good to me! could you put the generated files in the delphi folder in the gitattributes so that they are disabled for the diff too? (it was helpful this time but best to disable github from diffing those by default)

Good shout, done that!

PR #202, opened 2025-05-29
Add friendly frog at end of generate script

### LINKED ISSUE #199: Add a friendly frog at the end of the generate script to tell you it has finished
We want an indication that the generate script has run,

Let's add an ascii frog like we do in the update leapfrog script in spectrum desktop

```
cat << "EOF"

   (o)--(o)
  /.______.\    ribbit
  \________/
 ./        \.
( .        , )
 \ \_\\//_/ /
  ~~  ~~  ~~

EOF
echo "Done"
```

Also turn back on the automated running of codegen script for science. We can turn it off on our machines in someway?

r-ash:
Closes #199

### DIFF
--- a/scripts/generate
+++ b/scripts/generate
@@ -23,3 +23,16 @@ fi
 echo "Generating C++..."
 
 cd cpp_generation; uv run src/main.py
+
+cat << "EOF"
+
+   (o)--(o)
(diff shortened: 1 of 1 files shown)


M-Kusumgar (approved):
very cute

PR #208, opened 2025-05-30
changed how the prep data scripts brings in CD4 eligibility to be con…

mwalte10:
…sistent with the adult model

Removed the get_oridinal function I had written in favor of the findInterval that Jeff uses. To do this, had to change a > to a >= in the code. Also removed some commented out code. 

Fixes issue raised here: https://github.com/mrc-ide/frogger/issues/207


r-ash (approved):
Looks great, thank you!

PR #209, opened 2025-05-30
Small input data fixes

### LINKED ISSUE #206: Small input data fixes
There are a couple of bits of input data for the paediatric model which have small issues we should fix

1. 2. We take 1 off `paed_art_elig_cd4` when we pass into the C++ so for CD4 percent > 30 it will be -1, is this right?  https://github.com/mrc-ide/frogger/blob/4f998a46fd5f98f9a531ece046e1251f7876de97/scripts/read_spectrum.R#L893
    1. We shouldn’t take 1 off on the way in, and make this consistent with the adult model too.

They should both be worked out with C++ like indexing, and not -1 on the way in.

hcArtStart and hcArtEligCd4 should be integers

Fix the cut off for PMTCT, and remove art_mtct code as this is not used anymore

Move statically defined parameters into the C++ as consts, e.g. `hcAgeCoarse`, `hcAgeCoarseCd4`, `hcCd4Transition`, `adultCd4Dist`, `fertMultByAge`, `fertMultOffArt`, `fertmultOnArt`

r-ash:
Closes #206 

This PR just does a bit of tidying, it might be easiest to look at this commit by commit as they are fairly separate

1. Turns `hcArtStart` and `hcArtEligCd4` into integers
2. Fix the cut off of PMTCT. The spectrum modvar has index 1 for perinatal, 2 for breastfeeding < 350 and 3 for breastfeeding >= 350. We want to pass this into leapfrog with CD4 categories '>500', '350-500', '250-349', '200-249', '100-199', '50-99', '<50'. Previously we were filling index 1:4 i.e. 500 - 200 from index 2 in spectrum and 5:7 i.e. 199 - 0 with index 3. It should be 1:2 and 3:7. I have updated this in data prep scripts
3. Moved statically defined parameters into C++ as state space consts. `hcAgeCoarse`, `hcAgeCoarseCd4`, `hcCd4Transition`, `adultCd4Dist` @mwalte10 can you check these are ok to move? We wouldn't want to move them if you foresee them ever changing e.g. passing in different values for different countries or passing in different values to explore different outcomes
4. Remove `incrate` input param, it wasn't being used in the model


inst/include/models/child_model_simulation.hpp
M-Kusumgar, line 1: im not sure about this approach, you are linking two independent things together i think:
1. first independent thing is our set of building blocks (the dp, ha, hc configs), these define the variables models can use
2. second independent thing is the set of models, these can use any combination of building blocks

state space variables fall into 1 in terms of our implementation, because they interact with variables in set 1, such as `hc_age_coarse` being an index into `on_art` which comes from the intermediate state

declaring these things in the model code means that if anyone else develops a model that wants to use the hc config then they wont have access to `hc_age_coarse` or all the other variables which are very linked to parameter/state dimensions for example

i think these should be in the code generation

M-Kusumgar (changes requested):

inst/include/models/child_model_simulation.hpp
```diff
@@ -47,6 +48,108 @@ struct ChildModelSimulation<Config> {
...
+    tensor(2, 6) = 0.00990099;
+    tensor(3, 6) = 0.00710071;
+    tensor(4, 6) = 0.04960496;
+    tensor(5, 6) = 0.931993199;
+
+    real_type sum_col1 = tensor(0,0) + tensor(1,0) + tensor(2,0) + tensor(3,0) + tensor(4,0);
+    tensor(5, 0) = 1.0 - sum_col1;
+
+    real_type sum_col7 = tensor(0,6) + tensor(1,6) + tensor(2,6) + tensor(3,6) + tensor(4,6);
+    tensor(5, 6) = 1.0 - sum_col7;
```
M-Kusumgar: also side note, why is `tensor(5, 0)` and `tensor(5, 6)` defined above only to be reassigned below?

inst/include/models/child_model_simulation.hpp
r-ash: I see what you're saying, I was expecting them to have limited scope so moving them into essentially private data for the model variant seems fine to me. Probably worth chatting through with Maggie what these are actually for and if they would be used outside of the HIV child model. I agree with what you're saying if these do need to be used elsewhere

r-ash: Thanks, chatted through this and yep let's move it into the state space

inst/include/models/child_model_simulation.hpp
r-ash: Removed now :D

inst/include/models/child_model_simulation.hpp
```diff
@@ -142,12 +146,12 @@ struct ChildModelSimulation<Config> {
           for (int cat = 0; cat < hcTT; ++cat) {
             n_hc.hc2_hiv_pop(hd_alt, cat, 0, s) += c_hc.hc1_hiv_pop(hd, cat, hc1_ageend, s) *
                                                    p_dp.survival_probability(hc2_agestart, s, t) *
-                                                   p_hc.hc_cd4_transition(hd_alt, hd);
+                                                   hc_cd4_transition[hd_alt][hd];
           }
           for (int dur = 0; dur < hTS; ++dur) {
             n_hc.hc2_art_pop(dur, hd_alt, 0, s) += c_hc.hc1_art_pop(dur, hd, hc1_ageend, s) *
                                                    p_dp.survival_probability(hc2_agestart, s, t) *
-                                                   p_hc.hc_cd4_transition(hd_alt, hd);
```
mwalte10, line 150: What is the difference between the round and square brackets?

mwalte10 (approved):
This looks fine to me, just the one question on the difference between bracket types

inst/include/models/child_model_simulation.hpp
r-ash: This is because these are now in the state space so they are `std::array` types, not `Eigen::Tensor` types. So just different indexing style based on the type

PR #210, opened 2025-05-30
Compile child model for DLL

r-ash:
Required for #200 but that only be closed once mapping is done in spectrum desktop. Builds on #209


r-ash:
Accidentally pushed to wrong branch here!

Issue #213, opened 2025-06-03
Fix hardcoding in the spectrum_inputs file


mwalte10:

r-ash:
Done in #214

PR #214, opened 2025-06-03
Remove fertility parameter hard coding

mwalte10:
Solves issue described here: https://github.com/mrc-ide/frogger/issues/213

Removed hardcoding from the input generation script for fertility related parameters. Also changed structure to match what is in the dp file.


inst/include/models/child_model_simulation.hpp
```diff
@@ -187,6 +187,7 @@ struct ChildModelSimulation<Config> {
       i_hc.nHIVcurr = 0.0;
       i_hc.nHIVlast = 0.0;
       i_hc.df = 0.0;
+      auto a_fert_idx = std::floor(a/5 - 3);
```
r-ash: Can this end up being less than 0 and then we're indexing below with it?

What is the `-3` here?

mwalte10: For this to be less than zero the age when people start becoming fertile would have to be <3, which I think is an edge case we can rule out. 

This is mapping single year age groups into 5 year age bands from 15-49, so if a = 17,  floor(a / 5 - 3) will give that you should use the zero index value. 3 could be changed to p_fertility_age_groups / 5 to control for situations where p_fertility_age_groups != 15.

Issue #104, opened 2024-08-29
Adult testing against Spectrum


mwalte10:
- Determine what tests need to be run
- Run against all Spectrum files
- Copy over leapfrog tests from test-spectrum.R into frogger

r-ash:
Have made a new ticket #363 to cover this

PR #223, opened 2025-06-09
calculate base-year births and deaths outputs

### LINKED ISSUE #89: Births and deaths in the base year

jeffeaton:
This closes #89. Worked out code with Rob Glaubius this weekend.

Two questions arising while implementing this:

1. I wasn't quite sure where to implement the code -- in `run_model()` or `run_model_from_state()`. I ended up in `run_model()` because this calculation relies on the HIV epidemic being at zero; it would not address intializing outputs from a state after HIV epidemic had begun.

2. It felt like the code should be somewhere in a function rather than scripted in `run_model()`.  But I wasn't quite sure where such a function should live. The `GeneralDemographicProjection` class relies on a concept of a current state and next state, whereas this initialization pertains only to a current state, so didn't seem like it would fit there.

This is not tested yet; tests against Spectrum DP file output will be addressed in #104. In the meantime, here's a quick code check that I am using:

```r
devtools::load_all(".")

pjnz <- "tests/testthat/testdata/bwa_aim-no-special-elig-numpmtct.PJNZ"

par <- prepare_leapfrog_parameters(pjnz)
sr <- eppasm::read_hivproj_output(pjnz)
out <- run_model(par)

out$births[1] - sr$births[1]
max(abs(out$p_total_pop_background_deaths[,,1] - sr$natdeaths[,,1]))
```

With output:
```
> out$births[1] - sr$births[1]
        1970 
2.600495e-05 
> max(abs(out$p_total_pop_background_deaths[,,1] - sr$natdeaths[,,1]))
[1] 0.00033628
```

_Edit 10 June:_ I think that this will require regeneration of test data because it changes the births and background deaths output in the base year (previously would have been all zero). Let me know how to go about that or feel free to add commits for that.

Closes #89


inst/include/frogger.hpp
```diff
@@ -30,7 +30,51 @@ struct Leapfrog {
     State initial_state = {};
     initial_state.reset();
     if constexpr (ModelVariant::run_demographic_projection) {
-      initial_state.dp.p_total_pop = pars.dp.base_pop;
+
+      // Initialise base year ouputs
+      const auto& p_dp = pars.dp;
+      auto& is_dp = initial_state.dp;
```
mwalte10: How do you feel about using base rather than initial?

mwalte10: I guess could get confusing with what is a state space and what is an input? although the prefix should still differentiate that.

inst/include/frogger.hpp
```diff
@@ -30,7 +30,51 @@ struct Leapfrog {
...
+
+      // Births in base year
+      is_dp.births = 0.0;
+      for (int af = 0; af < SS::p_fertility_age_groups; ++af) {
+	
+	// female population exposed to births: average of current age a population
+	// and current age a+1 population survived backwards one year = average of current
+	// and previous year population. (Does not account for any base year net migration)
+
+	const auto a = SS::p_idx_fertility_first + af;
```
mwalte10: Can  you just start the for loop at SS::p_idx_fertility_first rather than adding on af here?

inst/include/frogger.hpp
```diff
@@ -30,7 +30,51 @@ struct Leapfrog {
...
+	auto female_fertility_population_a = is_dp.p_total_pop(a, SS::FEMALE) + is_dp.p_total_pop(a+1, SS::FEMALE) / p_dp.survival_probability(a+1, SS::FEMALE, t0);
+	is_dp.births += female_fertility_population_a * 0.5 * p_dp.age_specific_fertility_rate(af, t0);
```
mwalte10: why 0.5?

inst/include/frogger.hpp
```diff
@@ -30,7 +30,51 @@ struct Leapfrog {
     State initial_state = {};
     initial_state.reset();
     if constexpr (ModelVariant::run_demographic_projection) {
-      initial_state.dp.p_total_pop = pars.dp.base_pop;
+
+      // Initialise base year ouputs
+      const auto& p_dp = pars.dp;
+      auto& is_dp = initial_state.dp;
+      int t0 = 0;
```
mwalte10: should this be set as a state space default rather than hardcoded here?

inst/include/frogger.hpp
```diff
@@ -30,7 +30,51 @@ struct Leapfrog {
...
+	is_dp.births += female_fertility_population_a * 0.5 * p_dp.age_specific_fertility_rate(af, t0);
+      }
+
+      // Deaths in base year
+      // Note: this calculation would **probably** be more accurate if it survived the
+      // population 1 year backwards before calculating deaths (discussed with Rob
+      // Glaubius 7 June 2025)
+      for (int g = 0; g < SS::NS; ++g) {
+	      
+	// (a) age 0 deaths
```
mwalte10: Whats the purpose of the A/B/C notation?

mwalte10 (changes requested):

mwalte10:
Agree it should go somewhere other than run_model(), maybe a new function that is something like set_baseline()?

r-ash:
Re where to put this, when do we want this to be run? Should it be run for any model in the first year of the projection? i.e. any time we project 1970 (except for Brunei when we project 1985) or do we only want to run this in the first year for models run from Spectrum?

inst/include/frogger.hpp
jeffeaton: Average of current year and previous year female population

inst/include/frogger.hpp
jeffeaton: Just to denote three cases or steps; can delete it if it does not help clarity. Or change it to `// Case 1: ...`, `// Case 2: ...`, etc.

jeffeaton:
> Re where to put this, when do we want this to be run? Should it be run for any model in the first year of the projection? i.e. any time we project 1970 (except for Brunei when we project 1985) or do we only want to run this in the first year for models run from Spectrum?

Probably any model in the first year of a projection. It would not be required for an EPP simulation, but it is a negligible computational cost.

What are the options / how would it be different?

jeffeaton:
> Agree it should go somewhere other than run_model(), maybe a new function that is something like set_baseline()?

Where should that code go? @r-ash

inst/include/frogger.hpp
jeffeaton: Had a similar question when I was writing it. @r-ash

inst/include/frogger.hpp
jeffeaton: @r-ash -- opinion?

inst/include/frogger.hpp
jeffeaton: That doesn't solve the issue because the fertility rates array is indexed from 0 (refers to `af`). So would need to subtract off that constant in the next line:

```
        is_dp.births += female_fertility_population_a * 0.5 * p_dp.age_specific_fertility_rate(af, t0);
```

inst/include/frogger.hpp
r-ash: We're not using base anymore in the state space, we opted for more specific names so no clash anymore. `initial` feels more descriptive to me than `base`, but happy to be overruled if `base` has some specific meaning that I'm not aware of?

r-ash:
> > Agree it should go somewhere other than run_model(), maybe a new function that is something like set_baseline()?
> 
> Where should that code go? @r-ash

I'm with Maggie, I think a new function `set_baseline`. I'd put it in a brand new file in the `models` directory. And then call that new function from `run_model`. We'll also want to export the function for people who use the `run_single_year` function so that they can set baseline if they want to. We'll want to expose that to the Delphi interface when we switch that over to projecting year by year. We can do that exporting in another PR, but we should make a ticket for it. The alternative is to have logic in `run_single_year` to detect whether this is the first year of the projection or not and then call `set_baseline` if it is the first year. But I think being explicit and having the caller manage that.

inst/include/frogger.hpp
r-ash: I don't feel too strongly either way. The 100% reason to put something in the state space is if it might vary between model variants. The other reason would be if we want to use it in multiple places. Personally I'd be fine with it here, we would just need to be aware that if we want to use this constant elsewhere we're not defining it multiple times. At that point we would want to lift it into the state space. I think Mantra would probably err on putting it in the state space to collect all the global constants together, which is another good reason. Downside to state space is it means you have to look in the config to see the value it has, and it needs to flow through the code gen. Keeping it here means it is very simple. If we do leave it here though we should mark it as `const`, `const int t0 = 0;`

r-ash:
> I think that this will require regeneration of test data because it changes the births and background deaths output in the base year (previously would have been all zero). Let me know how to go about that or feel free to add commits for that.

You can regen the test data by running the script `./scripts/create_test_data.R`

inst/include/frogger.hpp
r-ash: Sorry Maggie I think I was being a bit slow in previous response! base fits better what we are calling this here "base-year births and deaths outputs"

So I think we should settle on one or the other. Call this `initial_state` and the function `run_initial_year_calculation` or `base_state` and the function `run_base_year_calculations`?

PR #225, opened 2025-06-09
changed data prep to not used data.table functions

### LINKED ISSUE #224: Remove data table syntax in data prep script

mwalte10:
Closes #224, switched some data.table language to dplyr


jeffeaton (commented):
Thanks for this. Nice. Two questions: 

* Can we remove the other `data.table` uses at the same time in order to remove it from the `Suggests:`?
   * https://github.com/mrc-ide/frogger/blob/b9c9a6b395da585dfab1b31751f63495e7c2e86a/scripts/compare_delphi.R#L45
   * https://github.com/mrc-ide/frogger/blob/b9c9a6b395da585dfab1b31751f63495e7c2e86a/tests/testthat/test-child-model.R#L130-L131 (and elsewhere in these tests)
   * https://github.com/mrc-ide/frogger/blob/b9c9a6b395da585dfab1b31751f63495e7c2e86a/scripts/create_test_data.R#L10
   * https://github.com/mrc-ide/frogger/blob/b9c9a6b395da585dfab1b31751f63495e7c2e86a/DESCRIPTION#L42

* Reviewing this reminded me my question from when I was looking at this — why is the `read_spectrum.R` in `scripts/` rather than `R/`? Is that just because it's not ready to go into package R functions yet and needs to be tidied up for that purpose still? [not for this PR, just a question]

mwalte10:
@jeffeaton Went through and removed those mentions of data.table, I think just vestiges of the data.table being required in the data prep script.  @r-ash just to note this included one edit in your compare_delphi.R script, which I can't run (rooted at your downloads) so if you could make sure that is still working as intended? 

I think that read_spectrum.R can go in R? And it not being in there is just a by-product of me not knowing where it should be.

r-ash:
I agree `read_spectrum.R` can go into `R` directory. I'm sure that `compare_delphi.R` script will be fine, I was only sourcing `data.table` to use the code in `read_spectrum.R` so should be fine now that has been updated.

mwalte10:
Great just moved that into the R directory. Will merge once tests pass.

Issue #165, opened 2025-04-11
Tidy up PJNZ reading


r-ash:
This has potential to be a biggie...

We have lots of different ways of reading from PJNZ
* Using [eppasm](https://github.com/mrc-ide/eppasm)
* Using [SpectrumUtils](https://github.com/rlglaubius/SpectrumUtils)
* Using [specio](https://github.com/mrc-ide/specio)

And then custom stuff in this package, and in Naomi and probably in other places

All of this results in quite hard to follow code, and means we end up reading the PJNZ file multiple times. 

It would be nice to have a centralised PJNZ reading place we can use

r-ash:
Note comment here https://github.com/mrc-ide/frogger/pull/226#discussion_r2143124435 , let's tidy up the code when we do this work

PR #226, opened 2025-06-11
run_model_from_state argument renaming and better error handling

### LINKED ISSUE #222: start_from_year argument in `run_model_from_state()`
https://github.com/mrc-ide/frogger/blob/b9c9a6b395da585dfab1b31751f63495e7c2e86a/inst/include/frogger.hpp#L39-L64

I have a few questions on this function.

* Is the purpose of this function, and the reason that it is a distinct function from `run_model()`, to be able to run a multi-year model simulation that does not start from the projection base year?

* `start_from_year` is confusing variable name to me because it is inconsistent from how we refer to things elsewhere (the `from` confuses me). What about referring to this as `simulation_start_year` or something?

* From reviewing the function, it looks like it allows an arbitrary simulation start year (`start_from_year`), but all simulations run for the full projection duration as the loop ends with `opts.proj_time_steps`

* Variable `time_step` feels confusing (ambiguity with the internal HIV time steps); what about `projection_step`?  And change `proj_time_steps` to `proj_steps` or `projection_steps`?

r-ash:
Closes #222 

Might be easier to look at the diff here commit by commit, as there is a bit of churn. This PR will
1. Rename `start_from_year` to `simulation_start_year`
2. Rename `proj_time_steps` to `proj_steps` and `time_step` to `proj_step`
3. Raise an error if the user requests an output year earlier than the `simulation_start_year`, this would previously just return 0s for any of those time dimensions, I think new behaviour is a little easier to understand.
4. Add examples for the 3 run functions `run_model`, `run_model_from_state` and `run_model_single_year`
5. Fix some issues in R CMD check
   1. Don't export functions only called internally @mwalte10 we should go through `read_spectrum` and `spectrum_inputs_paeds` and double check we're exporting what we need
   2. Fix note about global variable bindings
   3. Qualify packages used
   4. Move `spectrum_inputs_paeds` from scripts into the package, we're using these functions in `read_spectrum`


R/run_model.R
```diff
@@ -63,14 +85,28 @@ run_model_from_state <- function(parameters,
 #'   feed the returned list into the next single year model run. In
 #'   contrast to [run_model_from_state()] the `p_total_pop` output will
 #'   have dimensions 81 x 2 not 81 x 2 x 61.
+#' @examples
+#' pjnz <- system.file(
+#'   "pjnz/bwa_aim-adult-art-no-special-elig_v6.13_2022-04-18.PJNZ",
+#'   package = "frogger", mustWork = TRUE)
+#' demp <- prepare_leapfrog_demp(pjnz)
+#' proj <- prepare_leapfrog_projp(pjnz)
+#' parameters <- c(demp, proj)
```
jeffeaton: We now have a wrapper function for this
```
parameters <- prepare_leapfrog_parameters(pjnz)
```

R/spectrum_inputs_paeds.R
```diff
@@ -1,28 +1,26 @@
...
 ###Paediatric input formatting
 ###############################################
-library(dplyr)
-
 format_pmtct <- function(pjnz){
   pmtct_arv <- dp_read_pmtct(pjnz)
 
   ## remove all totals
-  pmtct_arv <- pmtct_arv %>% data.frame() %>% filter(!row_number() %in% which(rownames(pmtct_arv) %in% c('pmtct_noprophylaxis_percent', 'pmtct_total_number',
+  pmtct_arv <- pmtct_arv %>% data.frame() %>% dplyr::filter(!dplyr::row_number() %in% which(rownames(pmtct_arv) %in% c('pmtct_noprophylaxis_percent', 'pmtct_total_number',
```
jeffeaton: 1) This line is really hard to to parse. Helpful to code with one function per line, e.g.:

```
pmtct_arv <- pmtct_arv %>% 
    data.frame() %>% 
    dplyr::filter(
        !dplyr::row_number() %in% grep(pattern = 'postnatal', rownames(pmtct_arv))
    )
```

2) I don't love the pattern of converting array to data frame and then filtering on down number here -- just a lot to figure out what the code is doing, and seems a bit fragile.

I would probably go with something like:
```
 pmtct_arv_labels <- c('pmtct_noprophylaxis_percent', 
                       'pmtct_total_number',
                       'pmtct_postnatal_noprophylaxis_percent', 
                       'pmtct_postnatal_total_number',
                       'pmtct_postnatal_total_number')

pmtct_arv_row_idx <- which(rownames(pmtct_arv) %in%  pmtct_arv_labels)
pmtct_art <- ptmct_arv[pmtct_arv_row_idx, ]
```

No need to make edits in this PR, but maybe a long-list issue for when we consolidate this input prep code.

jeffeaton (approved):
Thanks for working through this; this looks good. I've left a couple of optional comments. Probably for separate future PR.

R/spectrum_inputs_paeds.R
mwalte10: Noted, what is the best way to track this? Should I make an issue?

mwalte10:
Checked to confirm that we only need to export the prepare_hc_leapfrog_projp. This doesn't compile for me, so hard to check that the test data generation is working correctly. Copying the error message here: 

`ℹ Loading frogger
Exports from C:/Users/mwalters/frogger/src/frogger.cpp:
   std::vector<std::string> list_model_configurations()
   Rcpp::List run_base_model(const Rcpp::List parameters, const std::string configuration, const std::vector<int> output_years)
   Rcpp::List run_base_model_from_state(const Rcpp::List parameters, const std::string configuration, const Rcpp::List initial_state, int simulation_start_year, const std::vector<int> output_years)
   Rcpp::List run_base_model_single_year(const Rcpp::List parameters, const std::string configuration, const Rcpp::List initial_state, int simulation_start_year)

Exports from C:/Users/mwalters/frogger/src/test_helpers.cpp:
   Rcpp::List serialize_vector(const Rcpp::List data, const std::string path1, const std::string path2)
   Rcpp::List deserialize_vector(const std::string path1, const std::string path2)

C:/Users/mwalters/frogger/src/RcppExports.cpp updated.
C:/Users/mwalters/frogger/R/RcppExports.R updated.
ℹ Re-compiling frogger (debug build)
── R CMD INSTALL ───────────────
─  installing *source* package 'frogger' ...
   ** using staged installation
   
      **********************************************
      WARNING: this package has a configure script
            It probably needs manual configuration
      **********************************************
   
   
   ** libs
   using C++ compiler: 'G__~1.EXE (GCC) 12.2.0'
   g++ -std=gnu++17  -I"C:/PROGRA~1/R/R-43~1.1/include" -DNDEBUG -I../inst/include/ -std=c++20 -I'C:/Users/mwalters/AppData/Local/R/win-library/4.3/Rcpp/include' -I'C:/Users/mwalters/AppData/Local/R/win-library/4.3/RcppEigen/include' -g -Wall -Wextra -pedantic -Wmaybe-uninitialized -Wno-unused-parameter -Wno-cast-function-type -Wno-missing-field-initializers -O2  -I"C:/RBuildTools/4.3/x86_64-w64-mingw32.static.posix/include"     -g -Wall -Wextra -pedantic -Wmaybe-uninitialized -Wno-unused-parameter -Wno-cast-function-type -Wno-missing-field-initializers -O2 -Wno-ignored-attributes -Wno-deprecated-copy -UNDEBUG -Wall -pedantic -g -O0 -fdiagnostics-color=always -c RcppExports.cpp -o RcppExports.o
   g++ -std=gnu++17  -I"C:/PROGRA~1/R/R-43~1.1/include" -DNDEBUG -I../inst/include/ -std=c++20 -I'C:/Users/mwalters/AppData/Local/R/win-library/4.3/Rcpp/include' -I'C:/Users/mwalters/AppData/Local/R/win-library/4.3/RcppEigen/include' -g -Wall -Wextra -pedantic -Wmaybe-uninitialized -Wno-unused-parameter -Wno-cast-function-type -Wno-missing-field-initializers -O2  -I"C:/RBuildTools/4.3/x86_64-w64-mingw32.static.posix/include"     -g -Wall -Wextra -pedantic -Wmaybe-uninitialized -Wno-unused-parameter -Wno-cast-function-type -Wno-missing-field-initializers -O2 -Wno-ignored-attributes -Wno-deprecated-copy -UNDEBUG -Wall -pedantic -g -O0 -fdiagnostics-color=always -c frogger.cpp -o frogger.o
   In file included from frogger.cpp:6
   ../inst/include/frogger.hpp:11:10:fatal error: t: No such file or directory
      11 | #include <format>
         |          ^~~~~~~~
   compilation terminated.
   make: *** [C:/PROGRA~1/R/R-43~1.1/etc/x64/Makeconf:272: frogger.o] Error 1
   ERROR: compilation failed for package 'frogger'
─  removing 'C:/Users/mwalters/AppData/Local/Temp/Rtmpo5it3T/devtools_install_647431f175c/frogger'
Error in `(function (command = NULL, args = character(), error_on_status = TRUE, …`:
! System command 'Rcmd.exe' failed
---
Exit status: 1
stdout & stderr: <printed>
---
Type .Last.error to see the more details.`

R/spectrum_inputs_paeds.R
r-ash: Fixed up both of those, feels worth while doing a bit of tidying whilst we're at it.

r-ash: We have quite a generic issue for this already, I've added a comment there referencing this https://github.com/mrc-ide/frogger/issues/165#issuecomment-2969957474

r-ash:
> Checked to confirm that we only need to export the prepare_hc_leapfrog_projp. This doesn't compile for me, so hard to check that the test data generation is working correctly. Copying the error message here:

Chatted about the compile error and we'll fix this over teams

mwalte10 (commented):
Looks good, thanks for help fixing the compiling error. What do you think of proj_steps_all rather than proj_steps? The difference between proj_step and proj_steps would be the type of error that I spend an hour trying to figure out when its just a small typo

jeffeaton:
I agree about `proj_step` and `proj_steps` being risky. `proj_steps_all` doesn't resonate for me though.

Issue #123, opened 2025-01-31
Set up packaging structure


M-Kusumgar:
In #227 we moved R package into a subdirectory. 

What I'd like is to have the header only library available built and testable using CMake. I tried this and initialised it with [cmake-init](https://github.com/friendlyanon/cmake-init) which looked great but ran into problems with the R package building. When we run R package installation it copies the R package files into a temp directory and builds from there. It won't copy anything from above the R root directory, so I don't think we can have the header only library living outside the repo.

Possibly we can work something out with copying the header only library in, but that will have to run before the R package is built. We can't use a `configure` script for this because that is run after the package bits get moved into a temp dir.

So it's all a bit fiddly. 

What I really want from C++ code being separate is
1. We can run it and test it with ctest outside of wrappers
2. We can run clang format and other static analysis tools
3. We can have it usable by others who might want to extend the model in C++ and they could use CMake too

We could try just a standalone C++ CMake `.cpp` file which imports the header only library from `inst/include`. That would get us 1 and 2 from above I think. And really no 3 is probably quite unlikely.

PR #227, opened 2025-06-12
Move R package into a separate sub directory

r-ash:
This PR will
1. Move the R package into a subdirectory `R`
2. Add a new README at the top level and update the R specific README slightly

In doing this, I had a quick go at moving the C++ header library code into a standalone directory built with CMake. I got this working ok but actually linking it into R was a bit of a faff, so I paused on that for now. This goes a good chunk of the way there. If had this set up as a repo which would build with CMake, run clang-format, cppcheck and ctest. We can still do this, might be easier to do that but have it compile the code at `inst/include` as actually getting R to pick up files outside of the package directory is difficult! Something we can look at again later.

Partly addresses issue #123 but I need to tidy that up
Expecting after this we can (I will add tickets for these)
1. Add a CI build for the c interface (this could run formatting and cppcheck for us)
2. Move the standalone model tests stuff out of the R repo. This really exists to prove we can compile and run the model without R headers, but with the c interface now existing this might be obsolete. It would still be nice to have some ctest that runs the model from some data


r-package/README.Rmd
```diff
@@ -0,0 +1,153 @@
...
+  "frogger",
+  repos = c("https://mrc-ide.r-universe.dev", "https://cloud.r-project.org"))
+```
+
+You can install the development version of frogger from [GitHub](https://github.com/) with:
+
+``` r
+# install.packages("remotes")
+remotes::install_github("mrc-ide/frogger", subdir = "R")
+```
```
jeffeaton: Does this need to be updated to reflect the `r-package/` subdirectory?

r-package/vignettes/development.Rmd
```diff
@@ -22,7 +22,7 @@ The overall aim of this is to allow researchers to write these model extensions
...
+Leapfrog has a set of `ModelVariant`s which can be run. See [`r-package/int/include/model_variants.hpp`](https://github.com/mrc-ide/frogger/blob/main/r-package/int/include/model_variants.hpp) for details of these. Each model variant is a collection of boolean switches or enums. These are used to turn on or off different parts of the code when it is run, so that leapfrog can have different extensions and we can compose these together in any way we want. These model variants are evaluated at compile time. When you compile the code, there will be an instance in the compiled binary for each variant your code will call. This will cause the binary to be larger but we chose to do it this way because we wanted the speed from the compile-time plymorphism. We still need to benchmark this to see how much of a gain it actually gives.
```
jeffeaton: is there a typo in the URL? 
```
(https://github.com/mrc-ide/frogger/blob/main/r-package/int/include/model_variants.hpp)
```

Should `.../int/include/...` -> `.../inst/include/...`?

r-package/vignettes/development.Rmd
```diff
@@ -64,7 +64,7 @@ Leapfrog runs a top-level loop over the time step. At the end of each time step,
 run_model(data, parameters, 1970:2030, 10, 2030)
 ```
 
-This time output is managed by the state saver see [`inst/include/state_saver.hpp`](https://github.com/mrc-ide/frogger/blob/main/inst/include/state_saver.hpp)
+This time output is managed by the state saver see [`r-package/int/include/state_saver.hpp`](https://github.com/mrc-ide/frogger/blob/main/r-package/int/include/state_saver.hpp)
```
jeffeaton: same question: `int/` -> `inst/`?

r-package/vignettes/development.Rmd
```diff
@@ -75,7 +75,7 @@ Leapfrog uses code generation to write the code for wiring up the input and outp
 
 ### Input data
 
-Input data is generated from csv file in [`inst/include/model_input.hpp`](https://github.com/mrc-ide/frogger/blob/main/inst/include/model_input.hpp), the columns in this file are
+Input data is generated from csv file in [`r-package/int/include/model_input.hpp`](https://github.com/mrc-ide/frogger/blob/main/r-package/int/include/model_input.hpp), the columns in this file are
```
jeffeaton: and again `int/` -> `inst/`?

jeffeaton: (going to stop with comments now; looks persistent. Maybe it's a find-replace error, or maybe it's correct as is)

jeffeaton (commented):
Looks good; couple of small questions

r-package/README.Rmd
r-ash: Good catch, yes it does

r-package/vignettes/development.Rmd
r-ash: Yep! A big find and replace error, thanks for spotting, should be fixed now

jeffeaton (approved):
Thanks! Looks good to me.
Jeff

PR #230, opened 2025-06-16
Eigen to nda: epic base

### LINKED ISSUE #174: Switch out Eigen for some other tensor library
Is there a better library to support row and column major support?

And an actively supported library.

Maybe [nda](https://github.com/TRIQS/nda)? Or what is [tensorflow](https://www.tensorflow.org/api_docs/cc) or pytorch using in C++ for their tensors, how do they look, can we use them?

### LINKED ISSUE #218: Replace the custom serialization with hdf5
We use this to save data from R and read in data from python. We also read in data in standalone model in C++ in this custom format. And I also have saved data in this format from Spectrum, in the TGBFixedArray class

M-Kusumgar:
This is the base for a couple of PRs to get NDA into frogger and get rid of eigen. Main changes in this branch include
* working version of NDA with R
* better defaults for the json schema so you dont have to specify the type of parameter (like using "TM" or "TFS" or "scalar" everywhere) and also the default for any alias is the cpp variable name so you only have to specify an alias from another language if it is different
* code generation for nda
* @mwalte10 's index fixes to stop out of bounds accessing
* removing initial state support temporarily
* removing serialisation support temporarily

Closes #174 
Closes #218


r-package/inst/include/models/child_model_simulation.hpp
```diff
@@ -206,10 +206,10 @@ struct ChildModelSimulation<Config> {
       i_hc.prev = i_hc.nHIVcurr / n_dp.p_total_pop(a + 15, 1);
 
       for (int hd = 0; hd < hDS; ++hd) {
-        i_hc.df += p_hc.local_adj_factor * p_hc.fert_mult_by_age(a_fert_idx) * p_hc.fert_mult_off_art(hd) *
+        i_hc.df += p_hc.local_adj_factor * p_hc.fert_mult_by_age(a_fert_idx, t) * p_hc.fert_mult_off_art(hd) *
                    (n_ha.h_hiv_adult(hd, a, 1) + c_ha.h_hiv_adult(hd, a, 1)) / 2;
         // women on ART less than 6 months use the off art fertility multiplier
-        i_hc.df += p_hc.local_adj_factor * p_hc.fert_mult_by_age(a_fert_idx) * p_hc.fert_mult_off_art(hd) *
+        i_hc.df += p_hc.local_adj_factor * p_hc.fert_mult_by_age(a_fert_idx, t) * p_hc.fert_mult_off_art(hd) *
```
M-Kusumgar, line 212: @mwalte10 is this okay? `fert_mult_by_age` was a 2d tensor but was being accessed as a 1d tensor, i put in time in the second dimension because that dimension was `proj_years` in the config

M-Kusumgar: also have ported over your index fixes in the rest of this file so would be worth having a skim to make sure i havent made any mistakes!

r-package/inst/include/options.hpp
```diff
@@ -1,6 +1,11 @@
 #pragma once
 
 #include <algorithm>
+#include <stdexcept>
+#include <string_view>
+#include <vector>
+#include <cmath>
+#include <string>
```
r-ash, line 8: Are these 3 imports used here? Did we just miss them off the includes before?

r-package/inst/include/serialize_eigen.hpp
```diff
@@ -1,168 +0,0 @@
-#pragma once
-
-#include <fstream>
-#include <iostream>
-#include <sstream>
-#include <string>
-#include <vector>
-
-#include <unsupported/Eigen/CXX11/Tensor>
```
r-ash, line 9: Are you planning to implement an equivalent for this with nda? Or try replace it with hdf5 as a PR into this epic?

r-package/src/frogger.cpp
```diff
@@ -3,8 +3,8 @@
 #include <string>
 #include <sstream>
 
-#include "frogger.hpp"
-#include "generated/r_interface/r_adapters.hpp"
```
r-ash: I think this shouldn't be needed with the Makevars set to `PKG_CPPFLAGS = -I../inst/include/` but does it give you better code completion or something?

r-package/inst/include/generated/config.hpp
```diff
@@ -16,53 +18,106 @@ struct DpConfig {
...
+    using shape_base_pop = nda::shape<
+      nda::dim<0, SS::pAG, 1>,
+      nda::dim<0, SS::NS, (SS::pAG)>
+    >;
+    nda::array_ref<real_type, shape_base_pop> base_pop;
+    using shape_survival_probability = nda::shape<
+      nda::dim<0, SS::pAG + 1, 1>,
+      nda::dim<0, SS::NS, (SS::pAG + 1)>,
+      nda::dim<0, nda::dynamic, (SS::pAG + 1) * (SS::NS)>
+    >;
```
r-ash, line 30: This does feel like a pretty ugly way to setup the shape. I get it gives you more flexibility but would be nice if they supported a simple initializer where you could just pass `SS::pAG + 1, SS::NS, nda::dynamic` and it could infer the stride and assume start at 0. I wonder if we could add a helper to make this a bit easier to parse?

r-ash (commented):
Just commenting for now as this is  WIP. But this all looks great. Couple of tiny comments but nothing substantial really.

Can we put a link to the library in the README, it is surprisingly hard to google when it is just called array :D https://github.com/dsharlet/array

r-ash:
I'm getting errors when compiling this on my windows machine. I'm curious if @mwalte10 you get the same? These are the same as we chatted about a bit ago Mantra.

Errors like
```
   C:\Users\Test\AppData\Local\Temp\cccFVq9v.s: Assembler messages:
   C:\Users\Test\AppData\Local\Temp\cccFVq9v.s: Fatal error: can't write 161 bytes to section .text of frogger.o: 'file too big'
   as: frogger.o: too many sections (36497)
   C:\Users\Test\AppData\Local\Temp\cccFVq9v.s: Fatal error: frogger.o: file too big
```

This is only when running `devtools::load_all()` with `devtools::check()` it compiles happily, as it does when compiling the C interface. 

Looks like this is because when running `devtools::load_all()` it passes `-g -O0` debugger flag. You can turn this off temporarily by running `withr::local_options(pkg.build_extra_flags = FALSE)`. It also works if you set flags `-Wa,-mbig-obj` in Makevars.

I think not an immediate issue for now but we should think about how to sort this out and document work around.

r-package/inst/include/generated/config.hpp
M-Kusumgar: music to my ears, basically i can make one, its pretty easy with template iteration, i was just cautious of adding in more template complexity into the codebase, since this is generated code it also doesnt need to be as readable but if you think its better to have a helper type, im more than happy to add it in! i wanna add it in too :P

r-package/inst/include/options.hpp
M-Kusumgar: yh i did get errors, i suspect they mightve been used in rcppeigen or something like that and were included by proxy

edit: apparently not, i tried deleting them and i only needed vector, have gotten rid of string and cmath, it was weird i was getting errors before

r-package/inst/include/serialize_eigen.hpp
M-Kusumgar: yh i thought no point getting that to work with nda when we are gonna change file formats, im adding in the file formats as part of this epic

r-package/src/frogger.cpp
M-Kusumgar: hmmm yh youre right, i think my lsp was complaining about it but ive just realised the inst/include location has changed, like the path relative to the root of the project because its in r-package! thats why i was getting those errors, ill revert this

r-package/inst/include/options.hpp
M-Kusumgar: edit: i did not lie! they are important! i was confused, its because the windows build with the c interface errors! thats why i added them in

r-package/inst/include/generated/config.hpp
M-Kusumgar: will add this helper type in another PR: https://github.com/mrc-ide/frogger/issues/237

PR #232, opened 2025-06-18
Add in support for initial state (for R)

M-Kusumgar:
unfortunately (or fortunately) nda doesnt have automatic conversion from "tensor map" (array_ref in nda) to "tensor" (array in nda). so we have to manually do the copy


r-package/inst/include/array/array.h
M-Kusumgar, line 1: this was a bug fix to the bug that paul and i found with this library, it affects how the compiler deciphers certain types with the iterators, see https://github.com/dsharlet/array/issues/109 for more details

r-ash (approved):
All looks great here too! I've updated the base branch

PR #233, opened 2025-06-18
Add support for c adapters

M-Kusumgar:
this just adds in support for c adapters, i havent added in support for the cpp adapters yet because reading in and reading out the data is disabled until hdf5 file formats are put in, ill just do that in the next PR


r-ash (approved):
All looking good here too

PR #234, opened 2025-06-19
Add hdf5 file format support

M-Kusumgar:
this adds in functions that read and save simple R lists to hdf5 format which all the tests now use

the child model has some other utils that are used which used to be saved in the same file as the parameters however after much fiddling around with them i settled on saving them to rds format because those utils dont need to be language agnostic and they have some weird R things going on with them

next up is probably re-enabling the standalone model stuff


r-package/scripts/create_test_data.R
```diff
@@ -15,17 +15,17 @@ pjnz_adult <- system.file("pjnz/bwa_aim-adult-art-no-special-elig_v6.13_2022-04-
 demp <- prepare_leapfrog_demp(pjnz_adult)
 proj <- prepare_leapfrog_projp(pjnz_adult)
 parameters <- c(demp, proj)
-saveRDS(parameters, testthat::test_path("testdata/adult_parms.rds"))
+save_hdf5_file(parameters, testthat::test_path("testdata/adult_parms.h5"))
```
r-ash, line 18: Do you want to reuse these exact files when running tests of Python interface? Just thinking losing the dimension labels in context of running tests might be a bit annoying. 

But I was thinking about removing these files altogether, and just creating the test data once in a `setup-` script. To try and avoid the churn in these files bloating git. Removing them won't be helpful for the Python interface though. Perhaps we should expect to keep these but look at using git LFS for them to avoid the bloat due to churn?

r-ash (approved):
This looks good, just some thoughts about the files we save out, but this looks good to go.

r-package/scripts/create_test_data.R
M-Kusumgar: i think if we need the labels of the dimensions we could do the child_utils pattern i did where there are extra RDS files specific to the R tests, i think the same files should be used for the python, R and c++ tests

yh thats tricky i was thinking about it too, maybe we should just delete these files and then tell people to run the create_test_data script in the readme? i guess they would need R installed but thats a dependency anyway

PR #235, opened 2025-06-26
Add standalone model back in

M-Kusumgar:
This adds the standalone model back into frogger. Main changes include:
* added an action for compiling the standalone model to check we havent broken that
* code gen for hdf5 utils and cpp adapters changed, the utils allow us to read and write to a hdf5 file and the cpp adapters use these
* have pulled out processing parameters from the run model functions because we need to save the hdf5 files with the c++ 0 based indexing (soon will switch out the names too after some discussion about it) 
* have added a helper to convert back to 1 based indexing `process_parameters_to_r`
* added a cmake build system to the standalone model
* updated readme and docs
* deleted test-standalone test in R, we will eventually write one that checked the hdf5 files from all the languages


.github/workflows/R-CMD-check.yaml
```diff
@@ -52,9 +52,6 @@ jobs:
           needs: check
           working-directory: r-package
 
-      - name: Extract test data
```
r-ash, line 55: I think I'd prefer to keep this in for now, unless we want to recreate the equivalent of the standalone test outside of R as the next PR.

Was this working as it was, or some additional work needed to make this test work again?

r-package/tests/testthat/test-standalone.R
```diff
@@ -1,129 +0,0 @@
-testthat::skip("Serialize tests disabled while we change serialization format to HDF5")
-
-test_that("We can compile the standalone program", {
```
r-ash, line 3: Earlier comment, I think I'd rather leave this test in, unless we're going to immediately add an integration test sitting outside of R which runs R and standalone model to check they are the same

r-package/scripts/create_test_data.R
```diff
@@ -120,7 +122,9 @@ out <- list(dp = dp,
             deaths_art = aids_deathsart,
             ctx_need = as.numeric(unlist(spec_ctx_need)))
 
-save_hdf5_file(c(proj, demp), testthat::test_path("testdata/child_parms.h5"))
+parameters <- c(proj, demp)
+parameters <- frogger::process_parameters_to_cpp(parameters)
+save_hdf5_file(parameters, testthat::test_path("testdata/child_parms.h5"))
```
r-ash: Worth using the wrapper here, and on line 19 above. `write_parameters` and `read_parameters`

I think I'd err on the side of not exporting `save_hdf5_file` and `read_hdf5_file`. We wouldn't expect people to be using those anymore?

r-package/inst/standalone_model/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cmake_minimum_required(VERSION 3.19)
+project(standalone_model)
+
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+find_package(HDF5 REQUIRED COMPONENTS C CXX)
+add_executable(simulate_model simulate_model.cpp)
+target_link_libraries(simulate_model HDF5::HDF5)
```
r-ash, line 9: Now this is a proper standalone thing, it would be nice to move this out of `inst` directory and into the root under `cpp_interface` or something

r-package/scripts/create_test_data.R
r-ash: If we do export them, would be nice to have some validation on the types of data that go in. Like check it is a list of named datasets/matrices and error otherwise

r-package/inst/standalone_model/README.md
```diff
@@ -10,97 +10,45 @@ program not from this CLI application. This standalone code just demonstrates
...
-devtools::install_local("../../")
-```
-
-Then run configuration from a shell using
+You will need hdf5 installed as this is the serialisation format we use.
+To install on linux run
 
 ```
-./configure
+sudo apt-get install libhdf5-dev
```
r-ash, line 21: We should probably think about windows people at some point, but no rush on it.

Would be nice to have some high level test that runs the model with R and standalone, and python in the future. Saves out to hdf5 and we can check same results are created for each model type. 

Probably extend this to a test that checks R data prep is equivalent to Python data prep (once we've added that many months in the future, if we ever get around to it)

cpp_generation/templates/cpp/cpp_interface/hdf5_utils.j2
```diff
@@ -0,0 +1,215 @@
+{% from "template_utils.j2" import comma_except_last %}
+{% include "header.j2" %}
+
+#include <filesystem>
+
+#include "H5Cpp.h"
```
r-ash, line 6: Good job working all this hd5 stuff out, looks quite tricky!

r-ash (approved):
It does feel a bit unfortunate we have to introduced the "OwnedPars", does add a fair bit of complexity compared to the magic Eigen was giving us. Suppose it is necessary complexity though!

.github/workflows/R-CMD-check.yaml
M-Kusumgar: i felt like the R test suite was the wrong place to put this test as it has nothing to do with R, i was just thinking we run a small script on ci maybe? i can just add a small one in the next pr where i also pull out the standalone model into a cpp folder?

r-package/tests/testthat/test-standalone.R
M-Kusumgar: yep we can have a small script that runs the standalone model and an R model and checks they produce the same output

r-package/scripts/create_test_data.R
M-Kusumgar: hmm yh with the wrappers, i say we only export `read_parameters` and `save_parameters` no point exporting the hdf5 utils anymore

r-package/inst/standalone_model/CMakeLists.txt
M-Kusumgar: yh can do it in the next pr, idm cpp_interface for now but i feel like we should standardise the naming, its like r-package, leapfrog-py, c_interface currently

r-package/inst/standalone_model/README.md
M-Kusumgar: yh i mean i can easily add a check next pr for r and standalone for the hivfullstrat model variant for now, we can do other model variants fairly easily with the standalone model too and same for python, data prep should be very easy too once its added

we can think about the windows people at some point too XD

cpp_generation/templates/cpp/cpp_interface/hdf5_utils.j2
M-Kusumgar: it was :'( also their docs arent great, at least their c++ docs

r-package/scripts/create_test_data.R
M-Kusumgar: okay there are a couple of files that come from `leapfrogR` in that script but just used `frogger:::save_hdf5_file` because they are not really parameters

.github/workflows/R-CMD-check.yaml
r-ash: Yeah that sounds fine. It is in the R tests because it was quick and easy to add there and then if it doesn't match, very easy to get a debugger in and look at the 2 data frames and see where they are different. Agreed it feels like it should be moved outside, especially as we are going to add the Python interface.

PR #236, opened 2025-06-27
Add model equality workflow

M-Kusumgar:
In response to the comments on the previous PR i have:
* pulled out the standalone model to the cpp_interface folder
* added a workflow for checking model equality between r and the standalone model outputs


r-package/scripts/create_model_output.R
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env Rscript
+root_dir <- here::here()
```
r-ash: So I think this is very similar to `run_model.R` script in `./r-package/scripts/run_model.R` can we consolidate these.

In fact I think that script will be broken since last PR as it was using standalone model data. 

Good to go with this, but can we add the `usage` here with docopt and let the caller pass in output dir.

r-ash (approved):
This looks good, small comment about tidying up existing `run_model.R` script

PR #239, opened 2025-07-14
Epic base for leapfrog-py

M-Kusumgar:
UV provides a nice template for building with scikit-build-core so i deleted the folder and remade it with
```
uv init --lib --build-backend scikit-build-core leapfrog-py
```
this uses pybind11 but nanobind is newer and made by the same author so i ported over the template to nanobind code. hopefully `uv build` should just work! more info in the readme :)


r-ash (approved):
Looks all good as a starting point for updating the python interface. Would be nice to keep some of the development info form the README. And I expect some of the pyproject stuff will need to come back once we get wheels building on CI. But not needed now

PR #240, opened 2025-07-18
Add python interface

M-Kusumgar:
This should enable basically everything, ive tested the adult model full strat and the child model and both have matched exactly between the python and the R output, feel free to have a play around!


r-package/R/hdf5_utils.R
```diff
@@ -29,7 +29,11 @@ save_datasets <- function(h5f, df, group = "") {
       h5f$create_group(new_path)
       save_datasets(h5f, dat, new_path)
     } else {
-      h5f[[new_path]] <- dat
+      if (is.logical(dat)) {
+        h5f[[new_path]] <- as.integer(dat)
+      } else {
+        h5f[[new_path]] <- dat
+      }
```
M-Kusumgar, line 36: this bug almost drove me crazy XD the R child model and python child model results didnt match, i had to call maggie to give me some likely culprits, narrowed it down to a couple and i checked and some of them were parsed incorrectly in the model code, but only a tiny bit wrong

the issue was booleans from R were being saved as uint8 in h5 format, python was reading them correctly, even c++ was reading the underlying data pointer correctly (since it was cast as a uint8 pointer), however putting the data vector into an `nda::array_ref` with data type int (32) meant that it wasnt read correctly, so printing the data pointer values worked in c++ and printing the same value in `Pars` struct showed the wrong result

the fix was pretty simple in the end, just convert it to a number

r-package/R/read_spectrum.R
```diff
@@ -796,7 +796,7 @@ prepare_hc_leapfrog_projp <- function(pjnz, params) {
   art$child_art[1,which(art$art_ispercent[2,])] <- art$child_art[1,which(art$art_ispercent[2,])] / 100
   v$paed_art_val <- art$child_art
   v$paed_art_age_spec <- art$age_spec
-  v$hc_art_start <- unname(which(colSums(art$child_art) > 0)[1]) - 1
+  v$hc_art_start <- as.integer(unname(which(colSums(art$child_art) > 0)[1]) - 1)
```
M-Kusumgar, line 799: this is an integer in the config, it was saved as a float in the h5 format which led to mismatched casting in the python interface, i guess r internals cast is giving us automatic conversion

r-package/R/hdf5_utils.R
r-ash: Nice spot!

cpp_generation/templates/cpp/config_mixer.j2
```diff
@@ -145,13 +146,13 @@ struct ConfigMixer<L, real_type1, ModelVariant1, Pair<true, {{ config.name }}Con
   };
 
   template<typename... Args>
-  static int build_output(int index, const OutputState& state, Args&&... args) {
+  static int build_output(int index, OutputState& state, Args&&... args) {
```
r-ash: Are we able to keep the `const` here? Any reason to remove that?

leapfrog-py/src/leapfrog_py/_core.pyi
```diff
@@ -1 +1,20 @@
-def return_20() -> int: ...
+def run_base_model(
```
r-ash, line 1: Excellent looking interface in the end!

r-ash (approved):
This all looks great, interface looks really clean. One tiny comment but this looks fine.

cpp_generation/templates/cpp/config_mixer.j2
M-Kusumgar: yep, i was getting a compilation error without const but could just simply change the type to const data in nanobind's ndarray and it works, have changed it

PR #241, opened 2025-07-22
Add python tests and interface

M-Kusumgar:
This pr adds an actual interface around the c++ model functions and adds tests for the python interface including test to see model output inequality between languages

i think code cov is a bit overkill for a thin interface

next up will probably add some docs!

### DIFF
--- a/leapfrog-py/pyproject.toml
+++ b/leapfrog-py/pyproject.toml
@@ -8,6 +8,7 @@ authors = [
 ]
 requires-python = ">=3.8"
 dependencies = [
+    "deepdiff>=8.4.2",
     "h5py>=3.11.0",
     "numpy>=1.24.4",
 ]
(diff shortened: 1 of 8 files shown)


leapfrog-py/pyproject.toml
```diff
@@ -8,6 +8,7 @@ authors = [
 ]
 requires-python = ">=3.8"
 dependencies = [
+    "deepdiff>=8.4.2",
```
r-ash, line 11: Can this go into the dev dependencies?

r-ash (approved):
This looks really good, very clean, I like all the types. One very minor comment but this looks good to go

leapfrog-py/pyproject.toml
r-ash: Oh I see you've moved that around in the next PR!

PR #243, opened 2025-07-23
Remove std format and get wheels building

M-Kusumgar:
I have removed std format and updates the build and publish to pypi action, the wheels will automatically build when commit is pushed to main and it will only upload wheels and sdist when we push a tag which also gives us the flexibility to push from a branch if we wish to

i did a run with a tag you can check out the:
* ci workflow: https://github.com/mrc-ide/frogger/actions/runs/16474262959
* published package on test pypi: https://test.pypi.org/project/leapfrog-py

### DIFF
--- a/.github/workflows/leapfrog-py-deploy.yaml
+++ b/.github/workflows/leapfrog-py-deploy.yaml
@@ -9,10 +9,6 @@ on:
     types:
       - published
 
-defaults:
-  run:
-    working-directory: leapfrog-py
-
 jobs:
   build_wheels:
     name: Build wheels on ${{ matrix.os }}
@@ -25,15 +21,19 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # Used to host cibuildwheel
+      - uses: actions/setup-python@v5
+
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==3.0.1
+
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.19.1
-        with:
-          package-dir: leapfrog-py
+        run: python -m cibuildwheel --output-dir wheelhouse leapfrog-py
 
       - uses: actions/upload-artifact@v4
         with:
           name: cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}
-          path: ./wheelhouse/*.whl
+          path: wheelhouse/*.whl
 
   build_sdist:
     name: Build source distribution
@@ -43,7 +43,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Build sdist
-        run: pipx run build --sdist
+        run: pipx run build --sdist leapfrog-py
 
       - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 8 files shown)


.github/workflows/leapfrog-py-deploy.yaml
```diff
@@ -29,10 +29,6 @@ jobs:
         uses: pypa/cibuildwheel@v2.19.1
         with:
           package-dir: leapfrog-py
-        env:
-          # Issue with compiling pybind11 eigen header as it has an invalid
-          # conversion from size_t to int
-          CIBW_SKIP: "pp38-manylinux_i686 pp39-manylinux_i686 pp310-manylinux_i686"
```
M-Kusumgar, line 35: no longer an issue :)

.github/workflows/leapfrog-py-deploy.yaml
```diff
@@ -3,15 +3,10 @@ name: Build and upload to PyPI
 on:
   workflow_dispatch:
   push:
+    tags:
+      - "v*"
```
r-ash, line 7: Not something we need to do now, but would like to think through the release process a bit more. Looks a little strange to see the git tag with v 0.1.0 when R package is a little behind it. I wonder if we have a single version number across the packages? What do you think? Would be nice to know when we're on R package 0.1.3 that should be the same underlying model as python 0.1.3. We would need to have some unified release process too with that, which might not be a bad thing!

cpp_generation/templates/cpp/c_interface/c_adapters.j2
```diff
@@ -19,12 +19,10 @@ template<typename T, size_t Rank>
...
   const auto size = shape.flat_max() + 1;
   if (length != size) {
-    throw std::invalid_argument(
-      std::format(
-        "Input data '{}' is the wrong size. Received array of length '{}', expected '{}'.",
-        name, length, size
-      )
-    );
+    std::ostringstream oss;
+    oss << "Input data '" << name << "' is the wrong size. "
```
r-ash, line 23: I'm sad to see the `format` go, ostringstream feels a bit dirtier https://youtu.be/AvqzGaZ-GoM

But I guess makes the building tidier :D

r-ash (approved):
Another great looking PR, nice work

.github/workflows/leapfrog-py-deploy.yaml
M-Kusumgar: yh i think this is sensible, we should have the same version across all our packages for sure, reduces confusion

cpp_generation/templates/cpp/c_interface/c_adapters.j2
M-Kusumgar: loooool love the video, yh we can add that in once c++ 26 puts in std format (im hoping)

.github/workflows/leapfrog-py-deploy.yaml
M-Kusumgar: ill do this version syncing in another pr: https://github.com/orgs/mrc-ide/projects/24?pane=issue&itemId=121305129&issue=mrc-ide%7Cfrogger%7C244

PR #256, opened 2025-08-07
Mkw issue 238

mwalte10:
This addresses issue #238

The stacked bar output is not output in the .dp file and needs to be manually extracted from the software each time the .pjnz file changes. For right now, I put those outputs in a .csv and load those into the test with a warning label that these must be regenerated each time the .pjnz file is changed.   

- [ ] I still need to add the new variables to the README, but I've completely lost where that is.


mwalte10:
C:/Users/mwalters/frogger/r-package/vignettes/variables.Rmd

PR #248, opened 2025-08-04
Jie patches

jeffeaton:

Issue #260, opened 2025-08-08
Demographic model misalignment


mwalte10:
In trying to work through the migration issue (#251 ), I noticed that the total population differs between Spectrum and Leapfrog. It differs both when migration is turned on and off. Can you remind me if this is expected? 

Code to see difference below (run from main): 

`
pjnz <-"/Data/Spectrum files/2025 final shared/ESA/Malawi_280525_FINAL.pjnz"
demp <- prepare_leapfrog_demp(pjnz)
proj <- prepare_leapfrog_projp(pjnz)
parameters <- c(proj, demp)
out <- run_model(parameters, "HivFullAgeStratification", 1970:2030)

dp <- eppasm::read_dp(pjnz)
spec_tot_pop <- data.table::data.table(SpectrumUtils::dp.output.bigpop(dp,direction = 'long'))
data.table::setnames(spec_tot_pop, 'Value', 'spec')
lfrog_tot_pop <- out$p_total_pop
dimnames(lfrog_tot_pop) <- list(Age = c(0:79, '80+'),
                                Sex = c('Male', 'Female'),
                                Year = 1970:2030)
lfrog_tot_pop <- data.table::data.table(reshape2::melt(lfrog_tot_pop))
lfrog_tot_pop <- lfrog_tot_pop[,.(Age = as.character(Age),
                                  Sex = as.character(Sex),
                                  Year = as.numeric(Year),
                                  lfrog = value)]
tot_pop <- merge(lfrog_tot_pop, spec_tot_pop, by = c('Age','Sex','Year'))
tot_pop[,diff := spec - lfrog]
`

jeffeaton:
I haven’t run the code below yet to see the nature and scale of the misalignment.  But what I have off the top of my head:

  *
There is a small difference I was aware of; I think I recall it was cropping up in an interaction of the HIV population and migration.
  *
There remain some slight differences in ART

If it is a difference in a file that is no migration and no ART, I’m not sure if I am expecting that.

I will run the example below and report back.

Thanks,
Jeff

From: mwalte10 ***@***.***>
Date: Friday, August 8, 2025 at 12:07 PM
To: mrc-ide/frogger ***@***.***>
Cc: Imai-Eaton, Jeffrey W ***@***.***>, Assign ***@***.***>
Subject: [mrc-ide/frogger] Demographic model misalignment (Issue #260)

[...]

Issue #251, opened 2025-08-05
net migration in child HIV model


jeffeaton:
@mwalte10 -- I don't see in the child HIV model where net migration is added/removed from the child HIV and ART populations (`hc1_` and `hc2_`).  Is that implemented?

For adult model, it is done here: https://github.com/mrc-ide/frogger/blob/17ad578d8020af6221140a4176ab4fa77c70f611/r-package/inst/include/models/hiv_demographic_projection.hpp#L82-L107

Overall, I think that child model should pass the test that `p_hiv_pop[1:15,,]` should aggregate to the same total as the `hc1_` and `hc2_` arrays, which it doesn't quite in the test file I am looking at.

Here's a test snippet illustrating:

```r
pjnz <- "~/Data/Spectrum files/2025 final shared/ESA/Malawi_280525_FINAL.pjnz"
par <- prepare_leapfrog_parameters(pjnz)
par <- prepare_hc_leapfrog_projp(pjnz, par)

lsim <- run_model(par, "ChildModel", 1970:2030)

p_hiv_child_total <- colSums(lsim$p_hiv_pop[1:15,,],,2)

hc_total <- colSums(lsim$hc1_hiv_pop,,4) + colSums(lsim$hc1_art_pop,,4) +
  colSums(lsim$hc2_hiv_pop,,4) + colSums(lsim$hc2_art_pop,,4)

cbind(p_hiv_child_total, hc_total)
```

```
> cbind(p_hiv_child_total, hc_total)
      p_hiv_child_total     hc_total
 [1,]           0.00000      0.00000
 [2,]           0.00000      0.00000
 [3,]           0.00000      0.00000
 [4,]           0.00000      0.00000
 [5,]           0.00000      0.00000
 [6,]           0.00000      0.00000
 [7,]          32.27178     32.25738
 [8,]          62.30089     62.22282
 [9,]         120.91275    120.62458
[10,]         233.45279    232.66855
[11,]         439.35084    437.56298
[12,]         798.68083    794.94785
[13,]        1401.51619   1394.17141
[14,]        2372.39748   2358.59315
[15,]        3880.04205   3855.32491
[16,]        6165.24452   6123.18253
[17,]        9437.92301   9374.42347
[18,]       13933.46429  13729.13618
[19,]       20444.21774  19793.46918
[20,]       28586.94738  27417.88888
[21,]       38160.32961  36751.43416
[22,]       48818.16089  47369.29745
[23,]       60492.32365  58852.27668
[...]

Issue #247, opened 2025-08-02
address case where there is 0 population but non-zero net migration


jeffeaton:
If there is zero population in an age group, the calculated net-migration rate will be `NaN`, resulting in an error in the population projection.

Commit [17ad578](https://github.com/mrc-ide/frogger/commit/17ad578d8020af6221140a4176ab4fa77c70f611) adds a check such that _if the net-migration is also zero_, the net-migration rate will be set to zero, avoiding numerical issue.

However, it is also plausible (though not likely to occur...) that there could be 0 population and non-zero net migration for an age group.  This situation will still lead to a `NaN` calculation in the population projection.

Not quite sure how to handle that. We should see what Spectrum does in that case.

Contrived example code:
```r
pjnz <- "~/Data/Spectrum files/2025 final shared/ESA/Malawi_280525_FINAL.pjnz"
parameters <- prepare_leapfrog_parameters(pjnz)
parameters <- prepare_hc_leapfrog_projp(pjnz, parameters)

## Set base population to zero for age 10
parameters$basepop[10,,1] <- 0.0

lsimP <- run_model(parameters, "DemographicProjection", 1970:2030)

lsimP$p_total_pop[1:15,,1]  # base year
lsimP$p_total_pop[1:15,,2]  # second year
```

Output:
```r
> lsimP$p_total_pop[1:15,,1]  # base year
        [,1]   [,2]
 [1,] 109827 110637
 [2,]  95511  97471
 [3,]  85567  88127
 [4,]  78723  81411
 [5,]  73863  76520
 [6,]  69914  72529
 [7,]  66845  69439
 [8,]  64295  66857
 [9,]  61567  64081
[10,]      0      0
[11,]  56729  59169
[12,]  54443  56822
[13,]  52154  54463
[14,]  49974  52209
[15,]  47907  50072
> lsimP$p_total_pop[1:15,,2]  # second year
           [,1]      [,2]
 [1,] 112053.16 112795.55
 [2,]  97989.85  99874.63
 [3,]  87854.24  90350.93
 [4,]  81306.34  83999.10
 [5,]  76140.77  78862.31
 [6,]  72210.38  74884.11
 [7,]  68742.53  71371.59
 [8,]  65968.63  68565.69
 [9,]  63614.44  66173.92
[10,]  61012.04  63524.45
[11,]       NaN       NaN
[12,]  56329.49  58743.15
[13,]  54088.76  56430.61
[14,]  51823.24  54101.45
[...]

PR #261, opened 2025-08-11
Mkw issue 251

mwalte10:
Addresses issue #251. Have implemented migration for the stratified paediatric population and tested using the following code: 

```
pjnz_child <-"C:/Users/mwalters/Imperial College London/HIV Inference Group - WP - Documents//Data/Spectrum files/2025 final shared/ESA/Malawi_280525_FINAL.pjnz"
demp <- prepare_leapfrog_demp(pjnz_child)
proj <- prepare_leapfrog_projp(pjnz_child)
proj <- prepare_hc_leapfrog_projp(pjnz_child, proj)
# parameters$netmigr_adj[] <- 0
out <- run_model(parameters, "ChildModel", 1970:2030)
hc1_hiv <- apply(out$hc1_hiv_pop, c(3,4,5), sum)
hc1_art <- apply(out$hc1_art_pop, c(3,4,5), sum)
hc1 <- hc1_hiv + hc1_art
dimnames(hc1) <- list(age = 0:4, sex = c('male','female'), year = 1970:2030)
p_hiv <- out$p_hiv_pop[1:5, , ]
dimnames(p_hiv) <- list(age = 0:4, sex = c('male','female'), year = 1970:2030)
hc1_df <- as.data.frame(as.table(hc1))
out_df <- as.data.frame(as.table(p_hiv))
colnames(hc1_df) <- c("Var1", "Var2", "Var3", "strat")
colnames(out_df) <- c("Var1", "Var2", "Var3", "pop")
c1 <- hc1_df %>%
  dplyr::inner_join(out_df, by = c("Var1", "Var2", "Var3")) %>%
  dplyr::mutate(diff = strat - pop)
c1 <- data.table::data.table(c1)
#expect_true(all(abs(c1$diff) < 1e-5))

hc2_hiv <- apply(out$hc2_hiv_pop, c(3,4,5), sum)
hc2_art <- apply(out$hc2_art_pop, c(3,4,5), sum)
hc2 <- hc2_hiv + hc2_art
dimnames(hc2) <- list(age = 5:14, sex = c('male','female'), year = 1970:2030)
p_hiv <- out$p_hiv_pop[6:15, , ]
dimnames(p_hiv) <- list(age = 5:14, sex = c('male','female'), year = 1970:2030)
hc2_df <- as.data.frame(as.table(hc2))
out_df <- as.data.frame(as.table(p_hiv))
colnames(hc2_df) <- c("Var1", "Var2", "Var3", "strat")
colnames(out_df) <- c("Var1", "Var2", "Var3", "pop")
c2 <- hc2_df %>%
  dplyr::inner_join(out_df, by = c("Var1", "Var2", "Var3")) %>%
  dplyr::mutate(diff = strat - pop)
c2 <- data.table::data.table(c2)
#expect_true(all(abs(c2$diff) < 1e-4))

all <- rbind(c1,c2)
all_strat <- data.table::copy(all)
all <- all[,.(strat = sum(strat), pop = sum(pop)), by = c('Var3')]
all[,diff := strat - pop]
all
all_strat[,year := as.integer(as.character(Var3))]
library(ggplot2)
ggplot(data = all_strat[], aes(year, diff, col = Var2, group = Var2)) + 
  geom_line() + facet_wrap(~Var1) + 
  labs(y = 'Difference between p_hiv_pop and hc1/hc2', col = 'Sex', x = NULL)
```

That results in the following differences: 
image

I've verified that the difference that arises at age 5 is due to the CD4 transition, happy to chat more about this @jeffeaton. 

One thing I'm not sure how to handle is when net migration is larger than the HIV positive population... Maybe would be good to attach to issue #247 ?

Other changes on this PR:

- I realized I was using the fertility start age (15) as the end age for children, but those are two different things that could differ so I've changed that.
- I realized some of the children state spaces had too many ages, so I've reduced those to the appropriate sizes. 
- The sex distributions of infections among one and two year olds uses the total population, which as we discussed in #260 diverges from Spectrum in some cases. I've changed it to use Spectrum's estimate of total population when the adult population output isn't being used.


mwalte10:
Added @r-ash and @M-Kusumgar as it looks like the tests that are failing are due to python code? Any suggestions here?

jeffeaton:
Thanks @mwalte10.  Regarding:
> One thing I'm not sure how to handle is when net migration is larger than the HIV positive population... Maybe would be good to attach to issue https://github.com/mrc-ide/frogger/issues/247 ?

I can see an issue if net migration is larger than the total population. I don't see why an issue would arise if net migration larger than the _HIV positive_ population. Can you clarify

jeffeaton:
> I realized I was using the fertility start age (15) as the end age for children, but those are two different things that could differ so I've changed that.

Ah, nice. I had also noticed that and addressed it in #248 (still in draft).  I think we can close #248 with this. I had used `p_idx_hiv_first_adult` as the end of the loop. But I think I like your approach better.

mwalte10:
> Thanks @mwalte10. Regarding:
> 
> > One thing I'm not sure how to handle is when net migration is larger than the HIV positive population... Maybe would be good to attach to issue #247 ?
> 
> I can see an issue if net migration is larger than the total population. I don't see why an issue would arise if net migration larger than the _HIV positive_ population. Can you clarify

You're right there wouldn't be a case when HIV pop is larger than total population so that actually covers that!

M-Kusumgar:
> Added @r-ash and @M-Kusumgar as it looks like the tests that are failing are due to python code? Any suggestions here?

i have managed to replicate it on my machine, will investigate this

cpp_generation/modelSchemas/configs/HcConfig.json
```diff
@@ -343,7 +347,7 @@
     },
     "hc_art_grad": {
       "num_type": "real_type",
-      "dims": ["SS::hDS", "SS::hcTT", "SS::hAG", "SS::NS"]
+      "dims": ["SS::hDS", "SS::hcTT",  "SS::hcAG_end", "SS::NS"]
```
M-Kusumgar: found the issue @mwalte10 ! i cant believe R tests passed and even python tests passed on windows and macos, this was a genuine and quite bad error

## Short explanation
if you look at the uses of `hc_art_grad` in the child model code then you can see we put the time index in the 1st position and disease stage (i think) in the 2nd position but here we've defined it the other way round, so to fix do this:
```json
"dims": ["SS::hcTT", "SS::hDS", "SS::hcAG_end", "SS::NS"]
```
thats the short version but you might be asking wtf? i didnt change any of the first two dimensions at all, i only changed the 3rd dimension so why is it breaking right now?

## Longer explanation
the way all of these tensor libraries work is that they have a one dimensional vector (basically flatten the tensor) and they index into it based on what what dimension indices you put in so if the size of the tensor is 7 x 4 x 66 x 2 (this tensor originally) and you want element with indices (5, 3, 12, 1) the flat index would be:
```
5 + 3 * (7) + 12 * (7 * 4) + 1 * (7 * 4 * 66) = 2210
```
(basically you take the product of previous dim sizes and multiply the index with it and sum). The maximum size of this tensor is simply 7 * 4 * 66 * 2 = 3696. Now our loop in the child model (inside `on_art_mortality`) indexes `hc_art_grad` with maximum indices of (2, 6, 14, 1) which corresponds to flat index
```
2 + 6 * (7) + 14 * (7 * 4) + 1 * (7 * 4 * 66) = 2284
```
which is < our max size 3696 so this worked in the original code. Let's take a look at what happens in the new code. We have max size as 7 * 4 * 15 * 2 (15 being `hcAG_end`) = 840 and our max index is still (2, 6, 14,1) which corresponds to a flat index of
```
    2 + 6 * (7) + 14 * (7 * 4) + 1 * (7 * 4 * 15) = 856
```
which is > 840! so we have an error there

r-ash: Nice spot!

cpp_generation/modelSchemas/configs/HcConfig.json
```diff
@@ -9,11 +9,11 @@
       "hc2DS": 6,
       "hc1_ageend": 4,
```
jeffeaton: Elsewhere in code, we have conventionally put upper boundaries as `< MAX` rather than `<= MAX`, as you have coded for `hcAG_end`.  Suggest revising this to 5 to be consistent.

r-package/R/read_spectrum.R
```diff
@@ -932,6 +932,11 @@ prepare_hc_leapfrog_projp <- function(pjnz, params) {
   total_births <- SpectrumUtils::dp.output.births(dp.raw = dp.x, direction = 'long')$Value %>% as.array()
   v$total_births <- total_births
 
+  bigpop <- SpectrumUtils::dp.output.bigpop(dp.raw = dp.x, direction = 'long')
+  bigpop <- array(bigpop$Value, dim = c(81, 2, 61), dimnames = list(age = c(0:79, '80+'),
+                                                          sex = c('Male', 'Female'),
+                                                          year = 1970:2030))
+  v$bigpop <- bigpop
```
jeffeaton: Suggest cutting this input down to what is needed for model calculation -- sex distribution in ages 2 and 3

r-package/inst/include/models/child_model_simulation.hpp
```diff
@@ -669,18 +669,29 @@ struct ChildModelSimulation<Config> {
...
-          total_pop_12_24 -= n_ha.p_infections(1, 0);
+         total_pop_12_24 -= n_ha.p_infections(1, 0);
         }
         auto uninfected_prop_12_24 = (n_dp.p_total_pop(1, s) - n_ha.p_hiv_pop(1, s)) / total_pop_12_24;
 
-        auto total_pop_24_plus = n_dp.p_total_pop(2, 0) - n_ha.p_hiv_pop(2, 0) +
-                                 n_dp.p_total_pop(2, 1) - n_ha.p_hiv_pop(2, 1);
         if (s == 1) {
-          total_pop_24_plus -= n_ha.p_infections(2, 0);
+         total_pop_24_plus -= n_ha.p_infections(2, 0);
```
jeffeaton: In this code, the hard coding of `1` and `2` suggest changing to some named const macro that describes what it is.

r-package/inst/include/models/hiv_demographic_projection.hpp
```diff
@@ -107,6 +109,62 @@ struct HivDemographicProjection<Config> {
...
+            for (int cat = 0; cat < hcTT; ++cat) {
+              n_hc.hc1_hiv_pop(hd, cat, a, s) *= 1.0 + migration_rate;
+            }
+            if (t >= p_hc.hc_art_start) {
+              for (int dur = 0; dur < hTS; ++dur) {
+                n_hc.hc1_art_pop(dur, hd, a, s) *= 1.0 + migration_rate;
+              }
+            }
+          }
+        }else{
```
jeffeaton: spaces: `} else {`

r-package/inst/include/models/hiv_demographic_projection.hpp
```diff
@@ -107,6 +109,62 @@ struct HivDemographicProjection<Config> {
...
+        real_type hc_migration_num = 0.0;
+        real_type hc_hivpop_postmig = n_ha.p_hiv_pop(a, s);
+        hc_migration_num = i_ha.hiv_net_migration(a, s);
+
+        real_type migration_rate = 0.0;
+        if (hc_hivpop_postmig > 0.0) {
+          migration_rate = hc_migration_num / (hc_hivpop_postmig - hc_migration_num);
+        }
+
+        if(a < hc2_agestart){
```
jeffeaton: spaces: `if (...) {`

r-package/inst/include/models/hiv_demographic_projection.hpp
```diff
@@ -36,6 +36,7 @@ struct HivDemographicProjection<Config> {
   static constexpr int PROJPERIOD_MIDYEAR = SS::PROJPERIOD_MIDYEAR;
```
jeffeaton, line 36: I think that the calculations done in `run_hc_hiv_pop_end_year_migration()` also need to be implemented where mid-year net migration is done.

Relevant code here: https://github.com/mrc-ide/frogger/blob/736257433520a41b9475658bd94337c320b44929/r-package/inst/include/models/hiv_demographic_projection.hpp#L310-L336

(That is obsolete in Spectrum now; but have it implemented in the adult model, so suggest keeping consistent here)

Note that there it will be `if (t = p_hc.hc_art_start)` rather than `if (t >= p_hc.hc_art_start)`

jeffeaton (changes requested):
Good stuff

r-package/R/read_spectrum.R
mwalte10: Done in commit 0429620

cpp_generation/modelSchemas/configs/HcConfig.json
mwalte10: done in commit 65479b3

r-package/inst/include/models/child_model_simulation.hpp
mwalte10: Have used enums to provide the following indices. These are what is relevant to this specific PR, but will ultimately aim to replace all hard coding with enums. I think the only others will be the PMTCT indicies. 

```
enum Index {
    age_0 = 0,
    age_1 = 1,
    age_2 = 2,
    sex_male = 0,
    sex_female = 1,
    tt_perinatal = 0,
    tt_bf05 = 1,
    tt_bf611 = 2,
    tt_bf1223 = 3,
    tt_bf2436 = 4
  };
```

r-package/inst/include/models/hiv_demographic_projection.hpp
mwalte10: fixed

r-package/inst/include/models/hiv_demographic_projection.hpp
mwalte10: fixed

mwalte10:
- I think all of @jeffeaton's comments have been addressed. 
- Not sure how to test the [end-year projection code](https://github.com/mrc-ide/frogger/blob/5f24b505535c9c17f067c42bad4dbf380201a24d/r-package/inst/include/models/hiv_demographic_projection.hpp#L341-L391)... suggestions? 
- I think we can improve on the function names in the hiv_demographic_projection file, have noted that on issue #258

r-package/inst/include/models/hiv_demographic_projection.hpp
mwalte10: done in commit [531a120](https://github.com/mrc-ide/frogger/pull/261/commits/531a120b14f90e78e0e4599036710f4042ee3672)

r-package/inst/include/models/child_model_simulation.hpp
mwalte10: Macro for male and female in the state space -> change to those

tt -> transmission timing
adopt Rob's naming and strata sent in the chat with the correct ranges. 

0to5m
6to11m
12to23m
24to35m

PR #262, opened 2025-08-11
pAG_INCIDPOP and pIDX_INCIDPOP as runtime option

### LINKED ISSUE #250: pAG_INCIDPOP is runtime option
https://github.com/mrc-ide/frogger/blob/17ad578d8020af6221140a4176ab4fa77c70f611/cpp_generation/modelSchemas/GlobalStateSpace.json#L10-L11

These two variables are run-time parameter values (see here: https://github.com/mrc-ide/eppasm/blob/a737e405379f477a39fa8f6e0d4cdbe2785ca917/src/eppasm.cpp#L184-L196). Therefore they can't be declared in `GlobalStateSpace.json` can they?

M-Kusumgar:
Closes #250 

basically just made 2 state space variables into parameters with a small renaming too


r-package/R/spectrum_inputs.R
```diff
@@ -282,5 +282,8 @@ prepare_leapfrog_projp <- function(pjnz, hiv_steps_per_year = 10L, hTS = 3) {
   v$cd4_mort_coarse <- projp$cd4_mort[ ,idx_expand_coarse, ]
   v$art_mort_coarse <- projp$artmx_multiplier * projp$art_mort[c(1, 2, rep(3, hTS - 2)), , idx_expand_coarse, ]
 
+  v$pAG_INCIDPOP <- 35L
+  v$pIDX_INCIDPOP <- 15L
+
```
M-Kusumgar: is this the best place to put this? @jeffeaton

r-ash (approved):
Looks good to me

r-package/R/spectrum_inputs.R
jeffeaton: The value for `pAG_INCIDPOP` is derived from the value of `v$incidpopage` defined here:

https://github.com/mrc-ide/frogger/blob/29f9aae5a0510d0890094be79cd9aad6463b83af/r-package/R/spectrum_inputs.R#L214-L217

I suggest moving the assignment of this to just below L217.  `pAG_INCIDPOP = 35` if `incidpopage = 0` and `pAG_INCIDPOP = 66` if `incidpopage = 1`.  Better for those values `35` and `66` to be assigned by macros rather than hard coded.

See here how it is handled in EPP-ASM (which is not ideal; better here): 

https://github.com/search?q=repo%3Amrc-ide%2Feppasm%20incidpopage&type=code

https://github.com/mrc-ide/eppasm/blob/a737e405379f477a39fa8f6e0d4cdbe2785ca917/R/read-spectrum-files.R#L1311-L1313

jeffeaton (changes requested):
Code changes look generally good; see comment response regarding assignment of the values. Let me know if you are happy to implement or if you would like me to.

r-package/R/spectrum_inputs.R
M-Kusumgar: sure and im not sure what you mean by macros in this context, or did you only mean macros in the eppasm c++ code? ive replaced the value of `pAG_INCIDPOP` with an if else on incidpopage for now

Issue #255, opened 2025-08-07
refactor run_wlhiv_births() to EPP-ASM expanding parameter inputs + indexing age by HIV age groups


jeffeaton:
In [`run_wlhiv_births()`](https://github.com/mrc-ide/frogger/blob/17ad578d8020af6221140a4176ab4fa77c70f611/r-package/inst/include/models/child_model_simulation.hpp#L176-L237), it currently follows the AIM paradigm of iterating through a collapsed array of parameter inputs by 5-year age group: 
https://github.com/mrc-ide/frogger/blob/17ad578d8020af6221140a4176ab4fa77c70f611/r-package/inst/include/models/child_model_simulation.hpp#L195
https://github.com/mrc-ide/frogger/blob/17ad578d8020af6221140a4176ab4fa77c70f611/r-package/inst/include/models/child_model_simulation.hpp#L209-L210

In EPP-ASM and elsewhere in the model we have taken the approach of expanding the parameter arrays to have the corresponding number of rows for the age iterator (e.g. for CD progression, ART mortality, etc.).

I suggest that we refactor this code to use the same approach for consistency and for smooth functioning when implementing the CD4-stratified FRR parameters — which will need to iterate along the collapsed HIV age groups (`ha`).  That becomes important here:
https://github.com/mrc-ide/frogger/blob/17ad578d8020af6221140a4176ab4fa77c70f611/r-package/inst/include/models/child_model_simulation.hpp#L209-L210

`n_ha.h_hiv_adult(hd, a, 1)` is indexed by `0` to `hAG`, not `0` to `pAG`.

For example of how this is handled in EPP-ASM:
```r
pjnz <- "~/Data/Spectrum files/2025 final shared/ESA/Malawi_280525_FINAL.pjnz"
fp <- eppasm::prepare_directincid(pjnz)
dim(fp$frr_cd4)
```
```
> dim(fp$frr_cd4)
[1]  7  8 61
```

Here 7 relates to HIV CD4 stages (`hTS`), 8 is the coarse age groups that are in 15-49 age range (`hAG`) and 61 is the project years.

Then it just iterates through that array; note the loop is over `ha` rather than `a`:
https://github.com/mrc-ide/eppasm/blob/a737e405379f477a39fa8f6e0d4cdbe2785ca917/src/eppasm.cpp#L858-L865

r-ash:
Example of current full vs coarse switching in the adult model

The "standard" input data used for full model
[...]

PR #263, opened 2025-08-12
Mkw issue 255

mwalte10:
To address issue #255 

Have changed data prep scripts to expand the fertility rate ratios from 5 year age bands into single year age groups and then have changed the indexing in the child model to reflect that. There was also a /0 happening in the adult model in the final simulation year that I have fixed here: https://github.com/mrc-ide/frogger/compare/main...mkw_issue_255#diff-ad6b8e10e8460cb29a153d2f6038d1a0e1d4bf9d867d40055907100e9f8d3fa3R361-R371

Would be useful to get feedback on:

1. Should the run_wlhiv_births model also be structured to use the coarse age distribution of the adults? And can you confirm that these are the correct age bands for the coarse population: 15-16, 17-19, 20-24, 24-29, 30-34, 35-39, 40-44, 45-49?
2. If the answer to 1 is yes, then that would involve changing the coarse fertility parameters to from 5 year age bands to the above groups. Is there code that does that for other parameters? Just want to replicate structure if so!


jeffeaton:
For expanding the FRRs to HIV age groups, you may be able to rely on existing output created by EPP-ASM: https://github.com/mrc-ide/eppasm/blob/master/R/spectrum.R#L187-L198

Try to mimic how it is handled for adult model age expansion: https://github.com/mrc-ide/frogger/blob/17ad578d8020af6221140a4176ab4fa77c70f611/r-package/R/spectrum_inputs.R#L275-L283

That might still be too obtuse... Let me know if you'd like me to look more carefully at what it does and explain better

mwalte10:
@M-Kusumgar Added you as this merges in the overwriting code.

PR #266, opened 2025-08-13
Overriding the intended way

### LINKED ISSUE #253: "overrides" only recognizes last "applies_if"
If multiple `"applies_if"` statements are specified in and `"overrides"` blog in a config, only the last `"applies_if"` is recognized.

For example:
```
    "overrides": [
      {
        "applies_if": "age105_open",
        "vars": {
          "pAG": 106
  }
      },
      {
        "applies_if": "age10_adult",
        "vars": {
          "p_idx_fertility_first": 10,
          "p_fertility_age_groups": 40
        }    
      }
    ]   
```

Comment from Mantra:
> just looked at this, youre right, the way the override stuff was thought about was we pick one of the overrides rather than do a merge of all the possible override combinations, but i think a merge can potentially be made possible

M-Kusumgar:
Closes #253 

This removes support for parameter overriding but adds support for flexible state space overriding. See #253 for more details.

This PR also removes coarse/full distinction between parameters and maps them to one name

I have also removed all the test data h5 and rds files, we now require people to generate these once at the start of development


cpp_generation/modelSchemas/configs/DpConfig.json
```diff
@@ -10,39 +10,37 @@
...
+    "age_specific_fertility_rate": {
+      "num_type": "real_type",
+      "dims": ["SS::p_fertility_age_groups", "opts.proj_steps"],
+      "alias": {
+        "r": "asfr"
+      }
+    },
+    "births_sex_prop": {
+      "num_type": "real_type",
+      "dims": ["SS::NS", "opts.proj_steps"]
```
mwalte10, line 43: I'm having trouble seeing what has changed here? Is this just a reordering?

r-package/R/spectrum_inputs.R
```diff
@@ -275,15 +276,17 @@ prepare_leapfrog_projp <- function(pjnz, hiv_steps_per_year = 10L, hTS = 3) {
...
+    v$cd4_initdist <- projp$cd4_initdist[ , idx_expand_coarse, ]
+    v$cd4_prog <- (1-exp(-projp$cd4_prog[ , idx_expand_coarse, ] / hiv_steps_per_year)) * hiv_steps_per_year
+    v$cd4_mort <- projp$cd4_mort[ ,idx_expand_coarse, ]
+    v$art_mort <- projp$artmx_multiplier * projp$art_mort[c(1, 2, rep(3, hTS - 2)), , idx_expand_coarse, ]
+  } else {
+    v$cd4_initdist <- projp$cd4_initdist[ , idx_expand_full, ]
+    v$cd4_prog <- (1-exp(-projp$cd4_prog[ , idx_expand_full, ] / hiv_steps_per_year)) * hiv_steps_per_year
+    v$cd4_mort <- projp$cd4_mort[ ,idx_expand_full, ]
+    v$art_mort <- projp$artmx_multiplier * projp$art_mort[c(1, 2, rep(3, hTS - 2)), , idx_expand_full, ]
+  }
```
mwalte10, line 289: I like this!

mwalte10: I feel like I remember you saying though that you wanted to avoid having to rerun parameter generation multiple times, did something change?

mwalte10 (commented):
This looks good to me. Just as a summary of how I interpret this changing the workflow: 
- No major change, but if there is a parameter that differs by age group (and thus could be different if the full vs coarse HIV model were run), that will now get set in the prepare_leapfrog_projp() function, with the name just indicating the parameter rather than specifying if it is coarse or full
- When prepare_leapfrog_projp() is run, there is a flag that indicates whether it should output the coarse or full parameters

Question: What result will you get if you run the full model with coarse parameters or vice versa? Is it interpretable?

r-package/scripts/create_test_data.R
```diff
@@ -122,7 +143,11 @@ out <- list(dp = dp,
             ctx_need = as.numeric(unlist(spec_ctx_need)))
 
 parameters <- c(proj, demp)
+parameters_coarse <- c(proj_coarse, demp)
+
 save_parameters(parameters, testthat::test_path("testdata/child_parms.h5"))
```
M-Kusumgar: child_parms_full instead

cpp_generation/modelSchemas/configs/DpConfig.json
M-Kusumgar: there is no more `"default": {...}` nesting in the object, its basically just the first line but i had to decrease the indentation of this whole block

r-package/R/spectrum_inputs.R
M-Kusumgar: yh definitely dont want to pollute the git repo with lots of big files, especially duplicates but ive just removed the problem from the git repo so all good now

cpp_generation/modelSchemas/GlobalStateSpace.json
```diff
@@ -1,11 +1,15 @@
...
+    "p_idx_hiv_first_adult": 15,
+    "hIDX_15PLUS": 0
+  },
+  "not_overridable": {
+    "MALE": 0,
+    "FEMALE": 1,
+    "ART0MOS": 0,
+    "PROJPERIOD_CALENDAR": 0,
+    "PROJPERIOD_MIDYEAR": 1
+  }
```
r-ash, line 14: Nice! protect the stuff

r-package/scripts/create_test_data.R
```diff
@@ -5,28 +5,45 @@
...
 
+devtools::load_all()
+
 # nolint start
-library(frogger)
 library(dplyr)
 
 ## Create demographic and projection parameters for adults
-pjnz_adult <- system.file("pjnz/bwa_aim-adult-art-no-special-elig_v6.13_2022-04-18.PJNZ", package = "frogger", mustWork = TRUE)
+pjnz_adult <- file.path("inst", "pjnz", "bwa_aim-adult-art-no-special-elig_v6.13_2022-04-18.PJNZ")
```
r-ash: This is v minor but if you use `here::here()` as the first arg to file.path this will then work if we run it from any dir within the package. At the moment it will only work if run from the root?

r-package/DESCRIPTION
```diff
@@ -42,7 +42,9 @@ Imports:
 Suggests:
     abind,
     docopt,
+    devtools,
     knitr,
+    leapfrog,
```
r-ash, line 47: We do want to be removing this dependency. I see we need it for generating the test data. Let's add a ticket for moving the testing away from referencing leapfrog, down the line these will become out of sync.

r-ash (approved):
This looks great, could we also remove all the old data test data form the history? Or just add this to a ticket to do at some point as part of some larger history butchery.

I like the separation of the test data, that makes the C++ much simpler.

I agree also with what Jeff said in out meeting yesterday. If that test data doesn't exist lets either skip or error with a nice message telling them what script they need to run to generate it. Alternatively, we could always generate this test data up front in a test helper, depends how slow it is?

cpp_generation/modelSchemas/GlobalStateSpace.json
M-Kusumgar: yep we should have some concepts of universal variables XD

r-package/scripts/create_test_data.R
M-Kusumgar: so i did this and then realised i dont do this for the child pjnz because its in testdata? like adult and child pjnzs are in different places, so instead ive moved the adult pjnz into the testdata dir and used test_path here too

r-package/DESCRIPTION
M-Kusumgar: yep have make an issue: https://github.com/mrc-ide/frogger/issues/268

M-Kusumgar:
> This looks great, could we also remove all the old data test data form the history? Or just add this to a ticket to do at some point as part of some larger history butchery.

yep have created a ticket with a scary sounding name: https://github.com/mrc-ide/frogger/issues/269 i was thinking when we replace leapfrog eventually we just replace its history with frogger but only one commit of frogger? like the latest commit? that way we have our history in frogger but actual leapfrog is light and portable

> I agree also with what Jeff said in out meeting yesterday. If that test data doesn't exist lets either skip or error with a nice message telling them what script they need to run to generate it. Alternatively, we could always generate this test data up front in a test helper, depends how slow it is?

its not super slow, less than 10 seconds for sure but if you havent compiled then devtools::load_all will also take time, i think since this is more of a one off thing, i would lean towards raising an error with a command instead of auto generating all the required files? ill add in a check, let me know what you think of the frog!

r-package/scripts/create_test_data.R
M-Kusumgar: loool so the examples in the documentation stopped working cos i moved the file out of inst/pjnz folder, i have instead put the child pjnz and xlsx files into the inst pjnz directory too

PR #278, opened 2025-09-05
[issue-277] Export HIV+ migrants from the model

### LINKED ISSUE #277: Export net HIV+ migrants from model
This is in intermediate data as `hiv_net_migration`. Does Spectrum need mid year or end year net migration?

r-ash:
Closes #277 

This moved `hiv_net_migration` from intermediate data into an output `p_hiv_net_migration` as we need it for Spectrum.

This also removes a testdata rds that was accidentally pushed. Not sure how that managed it as it looks like it is gitignored. I think it must have got committed before it got ignored, should be sorted now.

This also attempts to fix the pkgdown build failure on main that Maggie spotted. https://github.com/mrc-ide/frogger/actions/runs/17156044083/job/48673021196

The issue here is that it's trying to build the site from `/home/runner/work/frogger/frogger/docs` but it is actually in `/home/runner/work/frogger/frogger/r-package/docs` (since we moved the R package). Tried to fix this by appending the path with a `./`, if that doesn't work it'll work with `r-package/docs`.


cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -293,6 +289,10 @@
     "p_hiv_deaths": {
       "num_type": "real_type",
       "dims": ["SS::pAG", "SS::NS"]
+    },
+    "p_hiv_net_migration": {
+      "num_type": "real_type",
+      "dims": ["SS::pAG", "SS::NS"]
```
jeffeaton: Here I suggest name `p_hiv_pop_net_migration` for consistency with: https://github.com/mrc-ide/frogger/blob/12a4a8259ba077e24982671d9d14fb0a816fd101/cpp_generation/modelSchemas/configs/HaConfig.json#L261-L264

Note that `p_hiv_deaths` refers to something different because it is _HIV-related (caused) deaths_ in the (total) population, vs. `p_hiv_pop_net_migration` is the total migration among the HIV population.

I still prefer `p_hivpop_*` as I find it clearer grouping. The `p_hiv_pop_*` leaves a slightly awkward convention if we want to start outputting deaths by cause from the model:
`p_hiv_pop_total_deaths`
`p_hiv_pop_hiv_deaths` (weird)
`p_hiv_pop_background_deaths` 

 But I think I have lost that one.

jeffeaton (changes requested):
Thanks, this looks good. Suggestion on variable name.

cpp_generation/modelSchemas/configs/HaConfig.json
r-ash: Let's do `p_net_migration_hivpop` and in another PR switch the order of other names see 

https://teams.microsoft.com/l/message/19:35f85a8670f945f5ba74317f20393440@thread.tacv2/1757525621654?tenantId=2b897507-ee8c-4575-830b-4f8267c3d307&groupId=e24ac85e-dd9e-4505-8b9e-047b3a5fa6f0&parentMessageId=1757525621654&teamName=HIV%20Inference%20Group%20-%20WP&channelName=Leapfrog&createdTime=1757525621654

PR #281, opened 2025-10-10
Issue 275: Output ART and not on ART disaggregated non-AIDS deaths

### LINKED ISSUE #275: Output ART and not on ART disaggregated ~~excess~~ non-AIDS deaths
In Spectrum you can view non AIDS deaths. Under AIM -> Total Population -> Total non-AIDS deaths to HIV population, Total non-AIDS deaths to HIV population on ART, Total non-AIDS deaths to HIV population not on ART.

To fill this in in Spectrum we need this by time, sex, age (single year) and on ART/not on ART.

https://github.com/mrc-ide/frogger/blob/f257351916fc113d0696378bbadb68a647af7a17/r-package/inst/include/models/hiv_demographic_projection.hpp#L181 This is where it is output, but it is aggregated across on art / not on art. 

We need to output a stratified version of this

Alternatively we could approximate this posthoc (after model simulation) by dividing that total HIV deaths (by age / sex) by the ART coverage, since the mortality rate is same in both groups. That would obviate need for tracking additional stratified output.

r-ash:
Closes #275 

This PR will
1. Add a "Spectrum" model variant which runs post-hoc calculations to output additional indicators needed for Spectrum
2. Adds the first of these which is outputting on ART and not on ART disaggregated non-AIDS deaths
3. Also adds a note into the README about how to generate coarse stratified test data


r-package/inst/include/models/child_model_simulation.hpp
```diff
@@ -127,7 +127,6 @@ struct ChildModelSimulation<Config> {
 
   void run_child_model_simulation() {
     const auto& p_hc = pars.hc;
-    auto& n_hc = state_next.hc;
```
M-Kusumgar, line 130: nice! did your lsp tell you these are unused? what are you using?

r-package/README.md
```diff
@@ -19,14 +19,15 @@ You can install the development version of frogger from
 
 ``` r
 # install.packages("remotes")
-remotes::install_github("mrc-ide/frogger", subdir = "r-package")
+remotes::install_github("mrc-ide/frogger", subdir = "r-pacakge")
```
M-Kusumgar:
```suggestion
remotes::install_github("mrc-ide/frogger", subdir = "r-package")
```

M-Kusumgar (approved):
not sure i can check the validity of the actual for loops and maths calculations, but from the point of the setup around the model code, it all looks good from my end!

r-package/inst/include/models/child_model_simulation.hpp
r-ash: Just `devtools::load_all` told me this. They're not showing up for you? Maybe something I've got turned on in my Makevars

M-Kusumgar: yh not showing up for me! will look into this

r-package/inst/include/frogger.hpp
```diff
@@ -8,6 +8,7 @@
 #include "models/hiv_demographic_projection.hpp"
 #include "models/adult_hiv_model_simulation.hpp"
 #include "models/child_model_simulation.hpp"
+#include "models/spectrum_post_hoc_calculations.hpp"
```
jeffeaton, line 11: I would probably prefer `spectrum_posthoc_calculations`, but fine as is if you like better.

r-package/inst/include/models/spectrum_post_hoc_calculations.hpp
```diff
@@ -0,0 +1,101 @@
...
+  // private methods that we don't want people to call
+  private:
+  void calculate_nonaids_deaths() {
+    auto& n_ha = state_next.ha;
+    auto& n_sp = state_next.sp;
+    auto& i_sp = intermediate.sp;
+
+    for (int g = 0; g < NS; ++g) {
+
+      for (int ha = 0; ha < p_idx_hiv_first_adult; ++ha) {
```
jeffeaton: The looping in this code relies on pAG = hAG -- i.e. using the 'full' age stratification. I guess that will in practice be the case for this variant, but it doesn't necessarily need to be the case. The code could be written in such a way that it does a sensible allocation of coarse age to single age for the output.

Recommended changes: 

* Change the loop limit variables to correspond to the array dimensions. `h_hiv_adult` has an upper limit defined by `hAG`; use that in arithmetic for the loop rather than bounding on `pAG` (this pertains to line 76 more than here).
* Make the looping variable `a` if looping over all single-year ages (0...`pAG`); use `ha` if looping over adult HIV ages.

Optional change to consider: generalize the code so that it would also work if specified for the coarse age group HIV model. (On reading to the rest of the code, I think this is a very doable change; suggest it. Let me know if you'd like me to work on a PR for it.)

r-package/inst/include/models/spectrum_post_hoc_calculations.hpp
```diff
@@ -0,0 +1,101 @@
...
+        n_sp.p_deaths_nonaids_artpop(ha, g) = 0.0;
+        n_sp.p_deaths_nonaids_hivpop(ha, g) = 0.0;
+      }
+
+      for (int ha = p_idx_hiv_first_adult; ha < pAG; ++ha) {
+        i_sp.hiv_art_adult_sa = 0.0;
+        i_sp.hiv_untreated_adult_sa = 0.0;
+
+        for (int hm = 0; hm < hDS; ++hm) {
+          i_sp.hiv_untreated_adult_sa = i_sp.hiv_untreated_adult_sa + n_ha.h_hiv_adult(hm, ha - p_idx_hiv_first_adult, g);
```
jeffeaton: Use `+=` here? We use elsewhere in code and I think it's useful for readability.

jeffeaton (changes requested):
Thanks very much -- couple of comments.

r-package/inst/include/models/spectrum_post_hoc_calculations.hpp
r-ash: Thanks, good spot, I think I've sorted that so it should work with both now. Used the approach we have taken for this elsewhere.

r-ash:
Thanks Jeff, think that should all be fixed now.

PR #286, opened 2025-10-13
Age15 entrants refactor -- i_hc.age15_hiv_pop store adult CD4 distribution

jeffeaton:
This PR implements a suggested refactor of `i_hc.age15_hiv_pop` calculations such that the intermediate holds the number of entrants indexed by the _adult_ CD4 distribution, after applying the CD4 transition probabilities.

This benefits the virgin population implementation because it means this can be reused without needing to re-apply the CD4 transition probabilities (avoid code duplication), but I think it's a worthwhile change outside of that so implementing here as a separate PR. It also creates consistency with how age 15 entrants are passed in EPP-ASM (when not simulating child model).

A second small change is to rename `adult_cd4_dist` to `hc2_to_ha_cd4_transition` for clarity.

One minor thing not addressed:
* `age15_hiv_pop` is a bit of a misnomer for the age 10 variant of the model; we can just leave that for now and consider changing it later


cpp_generation/modelSchemas/configs/HcConfig.json
```diff
@@ -28,15 +28,15 @@
       "mtct_source": 11,
       "hc_age_coarse": [1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3],
       "hc_age_coarse_cd4": [0, 0, 0, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2],
-      "hc_cd4_transition": [
+      "hc1_to_hc2_cd4_transition": [
```
mwalte10, line 31: Nice, like this name. Same for adults.

r-package/inst/include/models/hiv_demographic_projection.hpp
```diff
@@ -200,21 +200,29 @@ struct HivDemographicProjection<Config> {
...
-      }
-    }
+      for (int hm = 0; hm < hDS; ++hm) {
+	i_hc.age15_hiv_pop(hm, g) = 0.0;
+	for (int hm_adol = 0; hm_adol < hc2DS; ++hm_adol) {
+	  auto age15_hivpop_hm_adol = 0.0;
+	  for (int htm = 0; htm < hcTT; ++htm) {
+	    age15_hivpop_hm_adol += c_hc.hc2_hiv_pop(hm_adol, htm, (hc2AG - 1), g);
+	  }
+	  i_hc.age15_hiv_pop(hm, g) += age15_hivpop_hm_adol * SS::hc2_to_ha_cd4_transition[hm][hm_adol];
```
mwalte10: Is SS: needed here? Can it be defined at the top of the function?

mwalte10 (approved):
Nice, one question on declaring SS:. I like the name changes and well noted on the misnomer of the age_15_entrants.

mwalte10 (commented):
Sorry had approved but meant to comment.  I like the name changes and one question on declaring SS:.

Noting that tests are failing... is this expected?

r-package/inst/include/models/hiv_demographic_projection.hpp
jeffeaton: To me it's clearer to have it here and be able to immediately see where that variable comes from, rather than having to flip back-and-forth to the top of the function to see where it is defined. (This flipping back and forth to find where things are defined was general point of reflection while developing the code.) 

I'm open to changing it though to be defined above.

jeffeaton:
> Sorry had approved but meant to comment. I like the name changes and one question on declaring SS:.
> 
> Noting that tests are failing... is this expected?

Thanks very much for quick review. I haven't looked at the source of test failures; in general I wouldn't have expected that. It should have been no change to simulation output, only calculation of that intermediate.

r-package/inst/include/models/hiv_demographic_projection.hpp
```diff
@@ -254,44 +262,23 @@ struct HivDemographicProjection<Config> {
...
+	  n_ha.h_hiv_adult(hm, 0, g) += i_hc.age15_hiv_pop(hm, g);
+	  if (t > opts.ts_art_start) {
+	    for (int hu = 0; hu < hTS; ++hu) {
+	      n_ha.h_art_adult(hu, hm, 0, g) += i_hc.age15_art_pop(hu, hm, g);
+	    }
+	  }
+	}
       }
     }
+    // TODO: implement static entrants to adult HIV population here for case when child model not simulated
```
r-ash, line 292: Can we add an issue number for this TODO?

r-ash:
> > Sorry had approved but meant to comment. I like the name changes and one question on declaring SS:.
> > Noting that tests are failing... is this expected?
> 
> Thanks very much for quick review. I haven't looked at the source of test failures; in general I wouldn't have expected that. It should have been no change to simulation output, only calculation of that intermediate.

It updates the intermediate `age15_hiv_pop` and `age15_art_pop` but then these are used in calculations for output `h_hiv_adult` and `h_art_adult`. It is the reference tests which check against current leapfrog implementation that are failing. Which makes sense if this change has not been updated in leapfrog too. Perhaps time to remove these tests if we're going to start diverging?

r-ash (commented):
Looks good, thanks for updating the variables vignette

r-ash:
The test `model agrees with leapfrog impl` in test-reference.R is failing

r-ash:
@jeffeaton to look at this one

jeffeaton:
> @jeffeaton to look at this one

Done -- good that I did! Indeed the reference test saved us here :/.

Addressed three things:

1. My refactor removed aging out from the youngest age group in the coarse HIV population.  
    - I think that maybe in the ChildModel variant, this was being skipped in the main branch code. This might be a source of some of the issues and discrepancies Maggie and I are finding (e.g. the AGO issue)
2. Addressed issue where if `hc_art_start` is before `ts_art_start`, children on ART were not aging into the adult HIV population. If this happens, they are added into the adult untreated HIV population. (I **think** this is how it is handled in Spectrum, but I can't remember for sure. It might come up later in reference testing.)
3. Added a test for ensuring that the HIV stratified population (`h_hivpop` + `h_artpop`) sums up to the total single-age HIV population (`p_hivpop`).  This would have caught the issue here [instead of the reference test] and caught some of the other issues that are cropping up.

r-ash (approved):
Thanks Jeff, this looks great, that test looks excellent.

Issue #299, opened 2025-11-04
declaration of scalars as intermediates -- why?


jeffeaton:
Do these scalars which are only used locally need to be declared in intermediates?
https://github.com/mrc-ide/frogger/blob/main/cpp_generation/modelSchemas/configs/HaConfig.json#L203-L212

These are simply used as temporary calculations in this loop and then discarded:
https://github.com/mrc-ide/frogger/blob/8cab8c45d957a6f32c81258b09b0db445207f08d/r-package/inst/include/models/adult_hiv_model_simulation.hpp#L125-L151

I see two reasons to save things in intermediates
1. Avoid re-allocations of heap memory blocks
2. Needs to be saved to be used in a different functions 

These don't tick either of those. And declaring it in intermediates metadata means that to edit the code I have to track up 3 directories and back down 4 into a different file.

Maybe there is a reason I am missing though, so not changing it now until consulting.

r-ash:
Agreed! I think this is just forgetfulness on my part, we should remove any scalars from the intermediate data where we can. I can update this.

jeffeaton:
Thanks very much.  If you have other things on your desk, I can also change these as I work through code updates and see places where they can be locally scoped.  Then later we can do a clean sweep if any left.

jeffeaton:
ChatGPT is working on this.

jeffeaton:
(This is not actually needed for release, but @r-ash  says we should "just do it")

PR #300, opened 2025-11-05
Issue 289 -- non-aids excess mortality

jeffeaton:
I did some refactor on the update for non-AIDS excess mortality. Three main changes

* The previous code only calculated non-AIDS excess mortality on the output; it needs to be part of the simulation code such that those are actual deaths removed from the population during simulation. The aspect that is calculated on outputs is the ART / non-ART split my single age.

* The ART / non-ART split calculation was updated so that it would theoretically work with coarse age groups (previously it

* In some cases I used `const auto` to declare temporary variables rather than declaring in intermediates -- see #299. Only did for new declarations, didn't change throughout.

* Changed a few parameter and state variable names for clarity (some work to do still to harmonize this).

I've made this PR to pull into your branch.


r-package/inst/include/models/spectrum_post_hoc_calculations.hpp
```diff
@@ -112,26 +112,34 @@ struct SpectrumPostHocCalculations<Config> {
 
       int a = p_idx_hiv_first_adult;
       for (int ha = 0; ha < hAG; ++ha) {
-        i_sp.excess_deaths_nonaids_no_art = 0.0;
-        i_sp.excess_deaths_nonaids_on_art = 0.0;
+
+	// Aggregate excess deaths and population in coarse age group
+        auto excess_deaths_nonaids_no_art_ha = 0.0;
+        auto excess_deaths_nonaids_on_art_ha = 0.0;
+	auto hivpop_ha = 0.0;
```
r-ash:
```suggestion
        // Aggregate excess deaths and population in coarse age group
        auto excess_deaths_nonaids_no_art_ha = 0.0;
        auto excess_deaths_nonaids_on_art_ha = 0.0;
        auto hivpop_ha = 0.0;
```

r-ash: Looks like some indenting is off? Unless it is the GitHub preview displaying it weirdly

r-ash: Actually just opened the actual file. Looks like some of the indents have been added with tabs instead of spaces causing the inconsistent indentation. Can you set your emacs to insert 2 spaces when you press tab? I think tabs can behave differently across editors.

r-ash (changes requested):
Thanks for fixing this up! That looks good. No real consequential comments, just a note about a mix of spaces and tabs for indentation which is making code harder to read I think.

> In some cases I used const auto to declare temporary variables rather than declaring in intermediates -- see https://github.com/mrc-ide/frogger/issues/299. Only did for new declarations, didn't change throughout.

Yes, I think I've been getting lazy with that. Totally agree with your comments in the issue, we should not use intermediate data for scalars unless they are needed elsewhere in the code. Keeping the scope to a minimum is always good.

The test failures are related to the new input for adult HIV model `cd4_nonaids_excess_mort` and `art_nonaids_excess_mort` in my PR I had added a separate prepare data function to export these. We just need to remove that function and move this code into the prep function for the adult model. Expect there will be a couple of tests to update too, to check that the new outputs are exported. Do you want me to do that?

r-package/inst/include/models/spectrum_post_hoc_calculations.hpp
r-ash: image

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -207,9 +225,12 @@
     "artpop_hahm": {
       "num_type": "real_type"
     },
-    "deaths": {
+    "deaths_hiv": {
```
mwalte10: This is just the aggregated number of HIV deaths by year right? Can it be calculated post hoc?

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -207,9 +225,12 @@
...
       "num_type": "real_type"
     },
+    "deaths_excess_nonaids": {
```
mwalte10: Same comment on if this can be calculated post hoc from h_deaths_excess_nonaids_no_art and h_deaths_excess_nonaids_on_art

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -290,6 +319,10 @@
       "num_type": "real_type",
       "dims": ["SS::pAG", "SS::NS"]
     },
+    "p_deaths_excess_nonaids": {
+      "num_type": "real_type",
+      "dims": ["SS::pAG", "SS::NS"]
+    },
```
mwalte10, line 319: Can children be exposed to non AIDS excess mortality? Or what is the reason for having both h_ and p_ outputs?

cpp_generation/modelSchemas/configs/HaConfig.json
mwalte10: Ah I see now that this is an intermediate

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -136,10 +137,15 @@ struct AdultHivModelSimulation<Config> {
...
-          i_ha.deaths = i_ha.cd4mx_scale * p_ha.cd4_mortality(hm, ha, g) * n_ha.h_hiv_adult(hm, ha, g);
-          i_ha.p_hiv_deaths_age_sex(ha, g) += opts.dt * i_ha.deaths;
-          n_ha.h_hiv_deaths_no_art(hm, ha, g) += opts.dt * i_ha.deaths;
-          i_ha.grad(hm, ha, g) = -i_ha.deaths;
+          i_ha.deaths_hiv = i_ha.cd4mx_scale * p_ha.cd4_mortality(hm, ha, g) * n_ha.h_hiv_adult(hm, ha, g);
+          i_ha.h_hiv_deaths_age_sex(ha, g) += opts.dt * i_ha.deaths_hiv;
+          n_ha.h_hiv_deaths_no_art(hm, ha, g) += opts.dt * i_ha.deaths_hiv;
+
+	  i_ha.deaths_excess_nonaids = p_ha.cd4_nonaids_excess_mort(hm, ha, g) * n_ha.h_hiv_adult(hm, ha, g);
+	  i_ha.h_deaths_excess_nonaids_agesex(ha, g) += opts.dt * i_ha.deaths_excess_nonaids;
```
mwalte10: I think would be best to align formatting between i_ha.h_deaths_excess_nonaids_agesex and  i_ha.h_hiv_deaths_age_sex. I think I prefer i_ha.h_excess_nonaids_deaths_age_sex and  i_ha.h_hiv_deaths_age_sex (mostly just on the ordering, don't have an opinion on age_sex vs agesex)

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -435,14 +447,22 @@ struct AdultHivModelSimulation<Config> {
       // remove hivdeaths proportionally to age-distribution within each age group
       a = p_idx_hiv_first_adult;
       for (int ha = 0; ha < hAG; ++ha) {
+
         if (i_ha.hivpop_ha(ha) > 0) {
-          i_ha.hivqx_ha = i_ha.p_hiv_deaths_age_sex(ha, g) / i_ha.hivpop_ha(ha);
+          i_ha.hivqx_ha = i_ha.h_hiv_deaths_age_sex(ha, g) / i_ha.hivpop_ha(ha);
+	  auto nonaids_excess_qx_ha = i_ha.h_deaths_excess_nonaids_agesex(ha, g) / i_ha.hivpop_ha(ha);
```
mwalte10: is this indexing intentional?

mwalte10 (commented):
Mostly just some suggestions about naming and indexing throughout

r-package/inst/include/models/spectrum_post_hoc_calculations.hpp
jeffeaton: > Actually just opened the actual file. Looks like some of the indents have been added with tabs instead of spaces causing the inconsistent indentation. Can you set your emacs to insert 2 spaces when you press tab? I think tabs can behave differently across editors.

Thanks -- sorry about that. Will try to figure how to fix my emacs. Sounds like something ChatGPT should be able to help with.

cpp_generation/modelSchemas/configs/HaConfig.json
jeffeaton: Have removed from intermediates per #299 discussion.

cpp_generation/modelSchemas/configs/HaConfig.json
jeffeaton: Have removed from intermediates per #299 discussion.

cpp_generation/modelSchemas/configs/HaConfig.json
jeffeaton: No children not currently exposed. The `p_` outputs provide output by single-year age. The `h_` outputs provide output by coarse age group (which is the scale on which they area calculated). It is different in the EPP-ASM variant of the model.

r-package/inst/include/models/adult_hiv_model_simulation.hpp
jeffeaton: The convention we agreed for the background deaths `deaths_background_totpop` would dictate I think `deaths_excess_nonaids_agesex`.

We also need to converge on "AIDS deaths" vs. "HIV deaths"

r-package/inst/include/models/adult_hiv_model_simulation.hpp
jeffeaton: can you clarify -- what indexing are you referring to?

r-package/inst/include/models/adult_hiv_model_simulation.hpp
mwalte10: Good point, agesex is good with me then. I think that AIDS deaths makes most sense as those deaths occur more proximally due to AIDS than HIV? Although that almost implies that we explicitly track AIDS cases...

r-package/inst/include/models/adult_hiv_model_simulation.hpp
mwalte10: Sorry, line 453 which is closer to the left margin than 451-452.

r-package/inst/include/models/adult_hiv_model_simulation.hpp
r-ash: Ticket here https://github.com/mrc-ide/frogger/issues/301

PR #304, opened 2025-11-10
Add new model variant for inputting transmission rate

r-ash:
I started looking at implementing the code for this and got confused quite quickly. I think @jeffeaton if you have capacity to translate the code over from EPPASM that would be much appreciated.

This adds the transmission rate input as a model variant, but possibly we could do this as a runtime parameter. I'm not sure at the moment what the best approach is. But if we get the code in place, we can then always move it from a variant into a runtime switch if we want to.


jeffeaton:
Closing this; addressed through runtime parameters instead of variant in #310

PR #306, opened 2025-11-12
incidence model variable renames

jeffeaton:
Rename variables in the incidence model to refer more intuitively to what they represent.


r-ash (approved):
Thanks, this looks good to me

PR #307, opened 2025-11-12
Incidence calc refactor

jeffeaton:
Is there a reason that it was preferable to split this code into two functions?

I've suggested refactoring to a single function for readability. The two bits do not do functionally distinct things and to me the multiple functions requires you to jump around to see what the code is doing.

Merge after #306


r-ash (approved):
Nope, I can't think of a good reason. I probably just split this when I was porting the code over initially but fine to consolidate that if it makes more sense for those things to be done together.

PR #310, opened 2025-11-14
Issue 303 JIE -- transmission rate model

jeffeaton:
This PR implements the EPP transmission rate model.

The tests are not perfect as they rely on ad hoc tolerance thresholds, but they will confirm the model runs and catch if something changes badly in an unexpected way.

One limitation about the tests is the current test data does not include adults on ART; tolerance thresholds likely will need to be revisited when ART added to test data.


cpp_generation/modelSchemas/GlobalStateSpace.json
```diff
@@ -10,6 +10,8 @@
     "FEMALE": 1,
     "ART0MOS": 0,
     "PROJPERIOD_CALENDAR": 0,
-    "PROJPERIOD_MIDYEAR": 1
+    "PROJPERIOD_MIDYEAR": 1,
+    "INCIDMOD_DIRECTINCID_HTS": 0,
+    "INCIDMOD_TRANSMRATE_HTS": 1
   }
```
jeffeaton, line 16: Is this the right place to specify these global constants? Or somewhere in the `HaConfig.json`

`GlobalStateSpace` feels like somewhat of a misnomer for this file if it is just specifying useful constants.

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -65,16 +65,32 @@ struct AdultHivModelSimulation<Config> {
...
       nda::fill(i_ha.h_hiv_deaths_age_sex, 0.0);
       nda::fill(i_ha.h_deaths_excess_nonaids_agesex, 0.0);
       run_disease_progression_and_mortality(hiv_step);
-      run_new_p_infections(hiv_step);
-      run_new_hiv_p_infections(hiv_step);
+
+      if (p_ha.incidence_model_choice == SS::INCIDMOD_DIRECTINCID_HTS) {
+	run_calc_new_infections_agesex(hiv_step);
+      } else if (p_ha.incidence_model_choice == SS::INCIDMOD_TRANSMRATE_HTS){
+	calc_new_infections_incidmod_transmrate(hiv_step);
```
jeffeaton: This feature switch could be here, or it could be inside a wrapper function `calc_new_infections_agesex()` that basically implements the feature switch. Any guidance or preference?

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -65,16 +65,32 @@ struct AdultHivModelSimulation<Config> {
...
+    }
 
     for (int hiv_step = 0; hiv_step < opts.hts_per_year; ++hiv_step) {
       nda::fill(i_ha.grad, 0.0);
       nda::fill(i_ha.gradART, 0.0);
       nda::fill(i_ha.h_hiv_deaths_age_sex, 0.0);
       nda::fill(i_ha.h_deaths_excess_nonaids_agesex, 0.0);
       run_disease_progression_and_mortality(hiv_step);
-      run_new_p_infections(hiv_step);
-      run_new_hiv_p_infections(hiv_step);
```
jeffeaton, line 85: You'll see I renamed these as `run_calc_new_infections_agesex()` and `run_add_new_hiv_infections()` to state more directly what the functions do. 

Question: is the prefix `run_` important / indicative of something here? Or could they simply be renamed `calc_new_infections_agesex()` and `add_new_hiv_infections()`?

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -320,6 +352,14 @@
...
       "num_type": "real_type",
       "dims": ["SS::pAG", "SS::NS"]
+    },
+    "prev15to49_hts": {
+       "num_type": "real_type",
+       "dims": ["opts.hts_per_year"]
+    },
+    "incid15to49_hts": {
+       "num_type": "real_type",
+       "dims": ["opts.hts_per_year"]
```
jeffeaton: These are outputs per HIV time step (10 per year). They are not used in Spectrum, etc., but useful to have for debugging and such.

In EPP-ASM, it just fills a vector of length years x steps; but that's not possible with the annual stepping paradigm here (I think). 

Does this seem to be a reasonable way to accomplish that output in the leapfrog annual step paradigm? Will output an array of years x steps dimension, which can be flattened.

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -8,26 +8,58 @@
...
+      "alias": {
+        "r": "iota"
+      }
+    },
+    "relative_infectiousness_art": {
+      "num_type": "real_type",
+      "alias": {
+        "r": "relinfectART"
+      }
+    },
```
jeffeaton, line 49: Per discussion yesterday, these are the new inputs which are option-specific. With this formulation, they will need to have an array there in all simulations, but it will just be ignored depending the feature flag. 

Seem okay? Any alternative approach suggested?

M-Kusumgar: this is probably the approach i wouldve taken too if you didnt want to declare a whole new model variant so makes sense to me!

cpp_generation/modelSchemas/configs/HaConfig.json
M-Kusumgar: yep sounds sensible to me

cpp_generation/modelSchemas/GlobalStateSpace.json
M-Kusumgar: so these constants get added to every single state space struct, basically the state space struct itself can change depending on model variant and these constants get added to all of them, happy to rename this to something that conveys that better

as for these new variables, if youre only using them in the Ha config then probably best to put them there, if you foresee any other variants/configs using these in the future you can leave them in here

r-package/inst/include/models/adult_hiv_model_simulation.hpp
M-Kusumgar: if its just gonna be in one place, i.e. here, i would prefer it to just be written out, i think over extracting everything into functions is also bad

M-Kusumgar (commented):
left some comments!

r-package/inst/include/models/adult_hiv_model_simulation.hpp
r-ash: Nothing special about `run_`. I probably used this because it's just a generic verb to use for a block of code which does some stuff. `calc` and `add` are perfectly good, no need to have the `run` also.

cpp_generation/modelSchemas/GlobalStateSpace.json
r-ash: Did we decide on putting everything const into state space? I have a vague recollection of talking about this but can't remember what we settled on. It could be useful to just have a pure C++ file `consts.cpp` which contains these. I think no reason they need flow through code gen and be in state space struct if they are a const and are not used for defining the size of any inputs or outputs.

cpp_generation/modelSchemas/configs/HaConfig.json
r-ash: Yep agreed this is the best for now, I've added a ticket for improving this in the future https://github.com/mrc-ide/frogger/issues/311

cpp_generation/modelSchemas/configs/HaConfig.json
jeffeaton: Thanks @M-Kusumgar , actually this doesn't work--I am getting a compiler error here:
```
   ../inst/include/generated/config.hpp:579:43: error: use of undeclared identifier 'opts'
     579 |       prev15to49_hts(shape_prev15to49_hts(opts.hts_per_year, output_years)),
         |                                           ^
   ../inst/include/generated/config.hpp:580:45: error: use of undeclared identifier 'opts'
     580 |       incid15to49_hts(shape_incid15to49_hts(opts.hts_per_year, output_years))
         |                                             ^
```

In the `parameters` block it does work to have dims that depend on `opts.`, e.g.: https://github.com/mrc-ide/frogger/blob/9af729c5e9b9b104851e87659319c66e3ec9d797/cpp_generation/modelSchemas/configs/HaConfig.json#L37-L42

Any tips?

jeffeaton: For now, I've hardcoded this as `"10"` to unblock

M-Kusumgar: with parameters this is okay because they are run time initialised and have dynamic sizes, with state objects, we only store arrays that we know the dimensions of at compile time, `opts.hts_per_year` is only known at runtime, i guess this raises the question if this is a compile time known quantity or not basically

jeffeaton: We conceptualized the code such that it could be changed at runtime if desired. 

In practice, this number is unlikely to change; can live with it if we need to shift to compile time parameter until there is a specific need for runtime variation

M-Kusumgar: i think that would be good, if it does change so infrequently, people can just edit the json where this can be stored as a statespace variable, regen cpp code and recompile but yh can also support it later as a runtime var

cpp_generation/modelSchemas/configs/HaConfig.json
mwalte10: Is it possible in this PR to remove the r aliases?

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -10,27 +10,50 @@
...
     },
     "overrides": [
       {
         "applies_if": "use_coarse_stratification",
         "vars": {
           "hAG_span": [2, 3, 5, 5, 5, 5, 5, 5, 31],
           "hAG": 9,
-          "h_fertility_age_groups": 8
+          "h_fertility_age_groups": 8,
+          "hAG_15to49": 8
```
mwalte10, line 27: Possible to do consistent naming convention here? I think I would be pro changing hAG_15to49 to h_15to49_age_groups

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -372,6 +395,18 @@
     },
     "hiv_births": {
       "num_type": "real_type"
+    },
+    "prevalence_15to49_hts": {
+       "num_type": "real_type",
+       "dims": ["10"]
```
mwalte10: Should these "10"'s be opts.hts_per_year?

r-package/tests/testthat/test-frogger.R
```diff
@@ -13,7 +13,7 @@ test_that("initial state set up works as expected", {
       "hiv_births_by_mat_age", "hiv_births",
       "p_deaths_background_totpop", "p_deaths_background_hivpop",
       "h_deaths_excess_nonaids_no_art", "h_deaths_excess_nonaids_on_art",
-      "p_deaths_excess_nonaids"
+      "p_deaths_excess_nonaids", "prevalence_15to49_hts", "incidence_15to49_hts", "artcoverage_15to49_hts"
```
jeffeaton, line 16: These tests with long strings of expected output names lists end up being pretty tedious to update. Any alternative approach here that only looks for critical things but doesn't care if other extraneous things come in?  Or saving the expected list once rather than repeating it in every test?

(Can be a created as a separate issue for future attention; just wanted to flag the question from working on them.)

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -68,21 +68,31 @@ struct AdultHivModelSimulation<Config> {
...
+
+      if (p_ha.incidence_model_choice == SS::INCIDMOD_DIRECTINCID_HTS) {
+	calc_new_infections_agesex(hiv_step);
+      } else if (p_ha.incidence_model_choice == SS::INCIDMOD_TRANSMRATE_HTS){
+	calc_new_infections_incidmod_transmrate(hiv_step);
+      } else {
+	throw std::invalid_argument("Incidence model choice not vaild\n");
+      }
```
mwalte10, line 94: Indentation a bit weird here

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -158,7 +168,108 @@ struct AdultHivModelSimulation<Config> {
...
+    real_type Xart = 0.0;
+
+
+    for(int s = 0; s < NS; ++s){
+      Xhivn_s[s] = 0.0;
+      Xhivn_incagerr[s] = 0.0;
+      for(int a = SS::pIDX_15to49; a < SS::pIDX_15to49 + SS::pAG_15to49; ++a) {
+	auto Xhivn_sa = n_dp.p_totpop(a, s) - n_ha.p_hivpop(a, s);
+	Xhivn_s[s] += Xhivn_sa; 
+	Xhivn_incagerr[s] += p_ha.incidence_rate_ratio_age(a - p_ha.pIDX_INCIDPOP, s, t) * Xhivn_sa;
```
mwalte10: Indentation also off here

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -158,7 +168,108 @@ struct AdultHivModelSimulation<Config> {
...
+	// calculation proportion of HIV population to include / exclude based
+	// on hivpop in single-year ages.
+	real_type prop_include;
+	if(ha == SS::hIDX_15to49){
+	  real_type hivp_ha = 0.0;
+	  int a = SS::pIDX_15to49;
+	  for(int i = 0; i < hAG_span[ha]; ++i, ++a) {
+	    hivp_ha += n_ha.p_hivpop(a, s);
+	  }
+	  prop_include = (hivp_ha > 0) ? 1.0 - n_ha.p_hivpop(SS::pIDX_15to49, s) / hivp_ha * (1.0 - opts.dt * hiv_step) : 1.0;
```
mwalte10: I feel like at some point we had decided to use the explicit if/else formatting for these types of things to help with readability

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -158,7 +168,108 @@ struct AdultHivModelSimulation<Config> {
...
+	Xhivn_incagerr[s] += p_ha.incidence_rate_ratio_age(a - p_ha.pIDX_INCIDPOP, s, t) * Xhivn_sa;
+      }
+
+      for(int ha = SS::hIDX_15to49; ha < SS::hIDX_15to49 + SS::hAG_15to49 + 1; ++ha){
+
+	// adjustment to first and last age group for partial year time step
+	// calculation proportion of HIV population to include / exclude based
+	// on hivpop in single-year ages.
+	real_type prop_include;
+	if(ha == SS::hIDX_15to49){
```
mwalte10: I think I do prefer declaring hIDX_15to49 = SS::hIDX_15to49 at the top, makes it a bit cleaner throughout

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -158,7 +168,108 @@ struct AdultHivModelSimulation<Config> {
...
+    const auto& p_ha = pars.ha;
+    auto& n_ha = state_next.ha;
+    auto& n_dp = state_next.dp;
+    auto& i_ha = intermediate.ha;
+    
+    // sum population sizes
+    real_type Xhivn_s[NS];
+    real_type Xhivn_incagerr[NS];
+    real_type Xhivp_noart = 0.0;
+    real_type Xart = 0.0;
```
mwalte10, line 182: What do the X's indicate?

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -158,7 +168,108 @@ struct AdultHivModelSimulation<Config> {
...
+    } // end loop over s
+
+  real_type Xhivn = Xhivn_s[MALE] + Xhivn_s[FEMALE];
+
+  // adjust HIV negative population for partial year time step
+  for(int s = 0; s < NS; s++){
+    Xhivn -= (n_dp.p_totpop(SS::pIDX_15to49, s) - n_ha.p_hivpop(SS::pIDX_15to49, s)) *
+      (1.0 - opts.dt * hiv_step);
+    Xhivn += (n_dp.p_totpop(SS::pIDX_15to49+SS::pAG_15to49, s) -
+	      n_ha.p_hivpop(SS::pIDX_15to49+SS::pAG_15to49, s)) *
```
mwalte10: Another case where it would be easier to read if these were defined above

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -158,7 +168,108 @@ struct AdultHivModelSimulation<Config> {
...
+      (1.0 - opts.dt * hiv_step);
+  }
+
+  real_type Xtot = Xhivn + Xhivp_noart + Xart;
+  real_type prevcurr = (Xhivp_noart + Xart) / Xtot;
+
+  int hts = t * opts.hts_per_year + hiv_step;
```
mwalte10: We moved t from ts correct? should hts then be ht?

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -372,6 +395,18 @@
...
+    },
+    "incidence_15to49_hts": {
+       "num_type": "real_type",
+       "dims": ["10"]
+    },
+    "artcoverage_15to49_hts": {
```
mwalte10, line 406: art_coverage_15to49_hts ?

r-package/inst/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -167,6 +278,10 @@ struct AdultHivModelSimulation<Config> {
     const auto adult_incid_first_age_group = p_ha.pIDX_INCIDPOP;
     const auto adult_incid_last_age_group = adult_incid_first_age_group + p_ha.pAG_INCIDPOP;
 
+    // Calculate HIV infections by age. This uses the updated
+    // 'current year' population [state_next] (vs. previous year
```
mwalte10, line 283: add note that previous year is state_curr... Also kind of confusing the then call state_next the current year.

mwalte10 (commented):
Some naming/ formatting suggestions.

r-package/tests/testthat/test-frogger.R
r-ash: Agreed with this! They are annoying. It is nice to have a test which checks that we are outputting everything we think we are. We could get that list from the config. This is the same cause of test failures as the Python ones. We could definitely remove this from there as we have a test which checks the output is the same across models across different architectures.

I wonder if we extend that common testing and move this into there and compare it to the config so then it will always be updated. I'll make a ticket.

r-ash: https://github.com/mrc-ide/frogger/issues/312

cpp_generation/modelSchemas/configs/HaConfig.json
```diff
@@ -10,27 +10,50 @@
       "hTS": 3,
       "hAG": 66,
       "p_fertility_age_groups": 35,
-      "h_fertility_age_groups": 35
+      "h_fertility_age_groups": 35,
+      "pIDX_15to49": 15,
+      "pAG_15to49": 35,
+      "pIDX_15to49": 15,
+      "pAG_15to49": 35,
```
r-ash: Duplicate keys here

cpp_generation/modelSchemas/configs/HaConfig.json
r-ash: Related to this discussion. https://github.com/mrc-ide/frogger/pull/310#discussion_r2527388030

`State` object has to be completely compile time known in size. So for now we could just make `hts_per_year` a state space variable. Alternatively if this really does need to be known at runtime we might need to make this as special as the "proj_years" and have a way to write these out into a proj_years x hts_per_year output state variable. That would probably require a fair bit of grossness in the config and generated code.

r-package/inst/include/models/adult_hiv_model_simulation.hpp
r-ash: Yeah I think there is a mix of tabs and spaces on these lines making the indentation look weird. We should always use spaces are they are more consistent across editors. Can you make your emacs insert spaces when you hit tab instead of an actual tab character?

r-package/tests/testthat/test-hiv-model.R
```diff
@@ -16,3 +16,238 @@ test_that("stratified HIV population equals single-age HIV population", {
...
+  ## Check that models produce non-zero epidemics
+  check_valid_0to1 <- function(x) {
+    all(!is.na(x)) &&
+      any(x > 0.0) && ## some non-zero
+      all(x >= 0) &&
+      all(x <= 1) 
+  }
+  
+  expect_true(check_valid_0to1(mod_full$prevalence_15to49_hts))
+  ## expect_true(check_valid_0to1(mod_full$artcoverage_15to49_hts))
```
r-ash, line 198: Do these commented out expects not pass when they should do?

r-ash (changes requested):
Looks good, couple of things I think we should resolve here. I'll fix up those python tests.

r-package/inst/include/models/adult_hiv_model_simulation.hpp
jeffeaton: Population size state variable accumulator (per comment on line 175)

r-package/inst/include/models/adult_hiv_model_simulation.hpp
jeffeaton: Indentation fixed

r-package/inst/include/models/adult_hiv_model_simulation.hpp
jeffeaton: For me, for a single use / minimally used macro, I prefer as here because the declarations:
1) Add extraneous lines to the top of the function that I have to wade through, but doesn't tell me how the variables are used.
2) Separates the information about where something comes from, so if I want to see where `hIDX_15to49` comes from I have to (1) scroll up to the top of the function to see it comes from `SS::`

r-package/inst/include/models/adult_hiv_model_simulation.hpp
jeffeaton: Can you have an `if {} else {}` on the right side of an assignment in C++?

r-package/inst/include/models/adult_hiv_model_simulation.hpp
jeffeaton: I renamed to `current_hiv_time_step` for clarity

r-package/inst/include/models/adult_hiv_model_simulation.hpp
jeffeaton: I agree about this confusion -- it was confusing to me when I was writing the code. Discuss in meeting tomorrow? Tagging @r-ash also

cpp_generation/modelSchemas/configs/HaConfig.json
jeffeaton: Thanks, good spot -- addressed

r-package/inst/include/models/adult_hiv_model_simulation.hpp
r-ash: No, you'd have to do

```
if (hivp_ha > 0) {
  prop_include = 1.0 - n_ha.p_hivpop....;
else {
  prop_include = 1.0;
}
```

M-Kusumgar:
lets chat about this in the leapfrog catch up but i think the different formatting comments do add noise to the PR review process, i think we shouldnt have to worry about this at all actually, i chatted with rob earlier but if people are happy to use clang format then we just have a standard formatter and you have to just run a command for it to format the codebase for you

and a ci job that'll check this and fail if you didnt run it

r-package/tests/testthat/test-hiv-model.R
jeffeaton: Yes; the ART coverage is 0 in all years because the test data don't have any ART.

cpp_generation/modelSchemas/GlobalStateSpace.json
r-ash: Let's keep these in json because then we can use them from R or python. Let's add an issue to expose these consts in other languages.

r-ash: Noting here also that you can override `ha` state space variables in `hc` but not the other way. However, you shouldn't do that and it is not intended behaviour.

cpp_generation/modelSchemas/configs/HaConfig.json
r-ash: `h_15to49_ag`?

r-ash: Let's go for `hAG_fertility` for now

r-package/inst/include/models/adult_hiv_model_simulation.hpp
r-ash: Let's just have a newline before the `?` and before the `:`

r-package/inst/include/models/adult_hiv_model_simulation.hpp
r-ash: As is for now, makes our abbreviations confusing. We can discuss at the next away day.

PR #317, opened 2026-01-26
Add function to export state space sizes for different model configurations

r-ash:
This PR will
1. Add a C++ function to return the state space dimensions `get_ss<ModelVariant>()`
2. Add function to R interface for calling this

I can add this to Python and other interfaces in a follow up PR, this looked like a nice amount to review to me.


M-Kusumgar (approved):
Nice that looks pretty neat to me!

PR #320, opened 2026-02-23
Update stale development doc to work with new model config setup

r-ash:
Updates info about input and output data config and how to add a new variant

### DIFF
--- a/leapfrogr/vignettes/development.Rmd
+++ b/leapfrogr/vignettes/development.Rmd
@@ -101,7 +101,57 @@ To add a new variant. Firstly, in the code generation:
 
 In the C++ code:
 
-1. Add new model code for your variant. At this point I would just add a skeleton with a print line to check your set up works, and fill in actual model code later. Model code should go [here](https://github.com/hivtools/leapfrog/tree/main/leapfrogr/inst/include/models), use one of the other models as an example of structure.
+1. Add new model code for your variant. At this point I would just add a skeleton with a print line to check your set up works, and fill in actual model code later. Model code should go [here](https://github.com/hivtools/leapfrog/tree/main/leapfrogr/inst/include/models). You can use the following snippet as a template
+   ```
+   #pragma once
+
+   #include "../options.hpp"
+   #include "../generated/config_mixer.hpp"
+
+   namespace leapfrog {
+   namespace internal {
+
+   // model_variant_flag1 & 2 need to be in pascal case, can do one of multiple model variant flags required for this model
+   template<typename Config>
+   concept {{ model_name }}Enabled = {{ model_variant_flag1 }}<Config> && {{ model_variant_flag2 }}<Config>;
+
+   template<typename Config>
+   struct {{ model_name }} {
+     {{ model_name }}(...) {};
+   };
+
+   template<{{ model_name }}Enabled Config>
+   struct {{ model_name }}<Config> {
+     using real_type = typename Config::real_type;
+     using ModelVariant = typename Config::ModelVariant;
+     using SS = Config::SS;
+     using Pars = Config::Pars;
+     using State = Config::State;
+     using Intermediate = Config::Intermediate;
+     using Args = Config::Args;
+
+     // function args
+     int t;
+     const Pars& pars;
+     const State& state_curr;
+     State& state_next;
+     Intermediate& intermediate;
+     const Options<real_type>& opts;
+
+     // only exposing the constructor and some methods
+     public:
+     {{ model_name }}(Args& args):
+       t(args.t),
+       pars(args.pars),
+       state_curr(args.state_curr),
+       state_next(args.state_next),
+       intermediate(args.intermediate),
+       opts(args.opts)
+     {};
+
+     void run() { std::cout << "Running new model\n" };
+   };
+   ```
 1. Call your new model variant at the appropriate point in the [`project_year` function](https://github.com/hivtools/leapfrog/blob/aa72a70ea4705489325185409223ddfc4aad75ef/leapfrogr/inst/include/leapfrog.hpp#L116)
 
 You now need to update the wrappers you want to expose the model variant to. For the R wrapper:
--- a/leapfrogr/vignettes/development.Rmd
+++ b/leapfrogr/vignettes/development.Rmd
@@ -22,7 +22,7 @@ The overall aim of this is to allow researchers to write these model extensions 
 
 ### Model variants
 
-Leapfrog has a set of `ModelVariant`s which can be run. See [`leapfrogr/inst/include/model_variants.hpp`](https://github.com/hivtools/leapfrog/blob/main/leapfrogr/inst/include/generated/model_variants.hpp) for details of these. Each model variant is a collection of boolean switches or enums. These are used to turn on or off different parts of the code when it is run, so that leapfrog can have different extensions and we can compose these together in any way we want. These model variants are evaluated at compile time. When you compile the code, there will be an instance in the compiled binary for each variant your code will call. This will cause the binary to be larger but we chose to do it this way because we wanted the speed from the compile-time plymorphism.
+Leapfrog has a set of `ModelVariant`s which can be run. See [`leapfrog/cpp_generation/modelSchemas/ModelVariants.json`](https://github.com/hivtools/leapfrog/blob/main/cpp_generation/modelSchemas/ModelVariants.json) for details of these. Each model variant is a collection of boolean switches or enums. These are used to turn on or off different parts of the code when it is run, so that leapfrog can have different extensions and we can compose these together in any way we want. These model variants are evaluated at compile time. When you compile the code, there will be an instance in the compiled binary for each variant your code will call. This will cause the binary to be larger but we chose to do it this way because we wanted the speed from the compile-time polymorphism.
 
 All model functions are templated on the model variant, and any conditional behaviour based upon the model variant should be written as an [`if constexpr`](https://www.learncpp.com/cpp-tutorial/constexpr-if-statements/).
 
@@ -58,7 +58,7 @@ Leapfrog runs a top-level loop over the time step. At the end of each time step,
 run_model(data, parameters, 1970:2030, 10, 2030)
 ```
 
-This time output is managed by the config see e.g. [`leapfrogr/inst/include/generated/config.hpp`](https://github.com/hivtools/leapfrog/blob/aa72a70ea4705489325185409223ddfc4aad75ef/leapfrogr/inst/include/models/adult_hiv_model_simulation.hpp#L110)
+This time output is managed by the internal `OutputState` struct see e.g. [`leapfrogr/inst/include/leapfrog.hpp`](https://github.com/hivtools/leapfrog/blob/aa72a70ea4705489325185409223ddfc4aad75ef/leapfrogr/inst/include/leapfrog.hpp#L110)
 
 ## Modifying config to add new input, output or intermediate data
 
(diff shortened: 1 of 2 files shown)


leapfrogr/vignettes/development.Rmd
```diff
@@ -22,32 +22,26 @@ The overall aim of this is to allow researchers to write these model extensions
...
+Leapfrog has a set of `ModelVariant`s which can be run. See [`leapfrogr/inst/include/model_variants.hpp`](https://github.com/hivtools/leapfrog/blob/main/leapfrogr/inst/include/generated/model_variants.hpp) for details of these. Each model variant is a collection of boolean switches or enums. These are used to turn on or off different parts of the code when it is run, so that leapfrog can have different extensions and we can compose these together in any way we want. These model variants are evaluated at compile time. When you compile the code, there will be an instance in the compiled binary for each variant your code will call. This will cause the binary to be larger but we chose to do it this way because we wanted the speed from the compile-time plymorphism.
```
M-Kusumgar: probably worth linking to the model variants json files instead? or are we not talking about the R code gen -> C++? probably worth mentioning that layer of indirection so they dont edit the model variants hpp file. I guess is this for researchers or software engineers? if its researchers then they will be editing the json file primarily.

also typo:

```suggestion
Leapfrog has a set of `ModelVariant`s which can be run. See [`leapfrogr/inst/include/model_variants.hpp`](https://github.com/hivtools/leapfrog/blob/main/leapfrogr/inst/include/generated/model_variants.hpp) for details of these. Each model variant is a collection of boolean switches or enums. These are used to turn on or off different parts of the code when it is run, so that leapfrog can have different extensions and we can compose these together in any way we want. These model variants are evaluated at compile time. When you compile the code, there will be an instance in the compiled binary for each variant your code will call. This will cause the binary to be larger but we chose to do it this way because we wanted the speed from the compile-time polymorphism.
```

leapfrogr/vignettes/development.Rmd
```diff
@@ -22,32 +22,26 @@ The overall aim of this is to allow researchers to write these model extensions
...
+   * `t` - the current time step of the model as an index, e.g. if running for 1970:2030, this will start at `1` and loop to `61`. Any input data you read based on time step index should have `1970` at the first index. Index `1` in R and `0` in C++.
+  * `pars` - the parameters for this model, these are read-only values
+  * `state_curr` - the state at the current point in time. This is read-only.
+  * `state_next` - the state at the next time point. This is what we are currently populating from the previous time step and the parameters.
+  * `intermediate` - a place for storing any data used within a single time step. Use this as to store intermediate values for use later in your code. This is reset to all 0s at the end of every time step.
+* A [public function](https://github.com/hivtools/leapfrog/blob/aa72a70ea4705489325185409223ddfc4aad75ef/leapfrogr/inst/include/models/adult_hiv_model_simulation.hpp#L64) that we can call from `project_year`
```
M-Kusumgar: some models do have two (if they need to run stuff at the end of a year for example)
```suggestion
* One or more [public functions](https://github.com/hivtools/leapfrog/blob/aa72a70ea4705489325185409223ddfc4aad75ef/leapfrogr/inst/include/models/adult_hiv_model_simulation.hpp#L64) that we can call from `project_year`
```

leapfrogr/vignettes/development.Rmd
```diff
@@ -64,82 +58,64 @@ Leapfrog runs a top-level loop over the time step. At the end of each time step,
 run_model(data, parameters, 1970:2030, 10, 2030)
 ```
 
-This time output is managed by the state saver see [`leapfrogr/inst/include/state_saver.hpp`](https://github.com/hivtools/leapfrog/blob/main/leapfrogr/inst/include/state_saver.hpp)
+This time output is managed by the config see e.g. [`leapfrogr/inst/include/generated/config.hpp`](https://github.com/hivtools/leapfrog/blob/aa72a70ea4705489325185409223ddfc4aad75ef/leapfrogr/inst/include/models/adult_hiv_model_simulation.hpp#L110)
```
M-Kusumgar: probably need to update this link

leapfrogr/vignettes/development.Rmd
```diff
@@ -64,82 +58,64 @@ Leapfrog runs a top-level loop over the time step. At the end of each time step,
...
+### Config structure
 
-### Input data
+The config is JSON, it has the following sections:
 
-Input data is generated from csv file in [`leapfrogr/inst/include/model_input.hpp`](https://github.com/hivtools/leapfrog/blob/main/leapfrogr/inst/include/model_input.hpp), the columns in this file are
+1. `name` - The name of this model variant type, it should be short. It is used in C++ code for the type which holds the state space, input data, intermediate data and output data associated with this variant.
+1. `long_name` - A long name for the model variant, at the moment used on for Delphi interface (to distinguish between 2 digit module codes used by Avenir internally)
+1. `namespace` - The name of an instance of the model variant, used in C++ code.
+1. `enable_if` - A conditional for when the model variant should be active. This is a compile time conditional. When the condition is true, the input data must be supplied and output data will be returned.
```
M-Kusumgar: worth mentioning this is one of the booleans in a model variant

leapfrogr/vignettes/development.Rmd
```diff
@@ -64,82 +58,64 @@ Leapfrog runs a top-level loop over the time step. At the end of each time step,
...
+1. `pars` - The inputs to the model used in this variant. You can use parameters supplied by other variants. Each parameter can define:
+   1. A "num_type" which should be "int" or "real_type". "real_type" is used by TMB but for normal running is just a `double`.
+   1. (optionally) "dims" - which is an array of sizes, it can use values from the state space, options or expressions e.g. `opts.proj_steps * opts.hts_per_year`. If no "dims" are set, assumed this is a scalar value.
+   1. (optionally) "alias" - a named list of aliases for the different language interfaces, at the moment only "r" is used and should be removed in the future. This should not be used for new parameters.
+1. `intermediate` - use to define any intermediate bits of data used during model fitting. We define them here because then they can be statically allocated instead of allocated every iteration of the time loop. These are automatically set to 0 at the end of every time step. Each piece of intermediate data needs a "num_type" and "dims".
```
M-Kusumgar: we arent fitting the model in leapfrog, this is the model that runs, fitting is all the extra stuff that happening in fitmod in eppasm for example

```suggestion
1. `intermediate` - use to define any intermediate bits of data used during the model run. We define them here because then they can be statically allocated instead of allocated every iteration of the time loop. These are automatically set to 0 at the end of every time step. Each piece of intermediate data needs a "num_type" and "dims".
```

leapfrogr/vignettes/development.Rmd
```diff
@@ -64,82 +58,64 @@ Leapfrog runs a top-level loop over the time step. At the end of each time step,
...
+1. `state` - data output from the model. This is the results filled in during a model fit. Each item needs a "num_type" and (optionally) "dims".
```
M-Kusumgar:
```suggestion
1. `state` - data output from the model. This is the results filled in during a model run. Each item needs a "num_type" and (optionally) "dims".
```
also worth mentioning that the output is an object/data frame/dictionary with keys of the state object and data equivalent to corresponding state object data with an extra time dimension

leapfrogr/vignettes/development.Rmd
```diff
@@ -64,82 +58,64 @@ Leapfrog runs a top-level loop over the time step. At the end of each time step,
...
 1. A flag which turns a section of code on or off, this will come with additional model inputs and outputs
 1. A flag which changes the dimensions of some model input or outputs
 
 The required changes will be different depending on if your new variant is one of the first types, which brings additional input/output or the second type which does not bring additional data.
 
 To add a new variant. Firstly, in the code generation:
 
-1. Add any new input or output data in [`leapfrogr/inst/include/model_input.csv`](https://github.com/hivtools/leapfrog/blob/main/inst/cpp_generation/model_input.csv) or [`leapfrogr/inst/include/model_output.csv`](https://github.com/hivtools/leapfrogr/blob/main/inst/cpp_generation/model_output.csv)
-1. If you have new input data, add a specialism for the new variant into the state_types template in [`inst/cpp_generation/state_types.hpp.in`](https://github.com/hivtools/leapfrog/blob/main/inst/cpp_generation/state_types.hpp.in)
+1. Add a new flag and variant in [`cpp_generation/modelSchemas/ModelVariants.json`](https://github.com/hivtools/leapfrog/blob/aa72a70ea4705489325185409223ddfc4aad75ef/cpp_generation/modelSchemas/ModelVariants.json)
```
M-Kusumgar: i think good to add something like "make sure the new flag is in all the other model variants"

M-Kusumgar (changes requested):
looks pretty good to me, i had a couple of comments but as well a those, i think it would be good to give someone a skeleton of model file when they are adding a new model to frogger, just something that gives them a start point (like the print message function you recommend), perhaps something like:

sorry couldnt find a nice syntax so just used jinja's `{{}}`

```cpp
#pragma once

#include "../options.hpp"
#include "../generated/config_mixer.hpp"

namespace leapfrog {
namespace internal {

// model_variant_flag1 & 2 need to be in pascal case, can do one of multiple model variant flags required for this model
template<typename Config>
concept {{ model_name }}Enabled = {{ model_variant_flag1 }}<Config> && {{ model_variant_flag2 }}<Config>;

template<typename Config>
struct {{ model_name }} {
  {{ model_name }}(...) {};
};

template<{{ model_name }}Enabled Config>
struct {{ model_name }}<Config> {
  using real_type = typename Config::real_type;
  using ModelVariant = typename Config::ModelVariant;
  using SS = Config::SS;
  using Pars = Config::Pars;
  using State = Config::State;
  using Intermediate = Config::Intermediate;
  using Args = Config::Args;

  // function args
  int t;
  const Pars& pars;
  const State& state_curr;
  State& state_next;
  Intermediate& intermediate;
  const Options<real_type>& opts;

  // only exposing the constructor and some methods
  public:
  {{ model_name }}(Args& args):
    t(args.t),
    pars(args.pars),
    state_curr(args.state_curr),
    state_next(args.state_next),
    intermediate(args.intermediate),
    opts(args.opts)
  {};

  void run() { std::cout << "Running new model\n" };
};
```
i do feel like this is probably one of the most confusing part of the step because theres concepts in there and template specialisation and stuff like that

leapfrogr/vignettes/development.Rmd
r-ash: Thanks, yeah that is just not right. Fixed that

leapfrogr/vignettes/development.Rmd
r-ash: Yes, for researchers, good shout linking to the doc they'll actually change.

r-ash:
> looks pretty good to me, i had a couple of comments but as well a those, i think it would be good to give someone a skeleton of model file when they are adding a new model to frogger, just something that gives them a start point (like the print message function you recommend), perhaps something like:
> 
> sorry couldnt find a nice syntax so just used jinja's `{{}}`
> 
> ```c++
> #pragma once
> 
> #include "../options.hpp"
> #include "../generated/config_mixer.hpp"
> 
> namespace leapfrog {
> namespace internal {
> 
> // model_variant_flag1 & 2 need to be in pascal case, can do one of multiple model variant flags required for this model
> template<typename Config>
> concept {{ model_name }}Enabled = {{ model_variant_flag1 }}<Config> && {{ model_variant_flag2 }}<Config>;
> 
> template<typename Config>
> struct {{ model_name }} {
>   {{ model_name }}(...) {};
> };
> 
> template<{{ model_name }}Enabled Config>
> struct {{ model_name }}<Config> {
>   using real_type = typename Config::real_type;
>   using ModelVariant = typename Config::ModelVariant;
>   using SS = Config::SS;
>   using Pars = Config::Pars;
>   using State = Config::State;
>   using Intermediate = Config::Intermediate;
>   using Args = Config::Args;
> 
>   // function args
>   int t;
>   const Pars& pars;
>   const State& state_curr;
>   State& state_next;
>   Intermediate& intermediate;
>   const Options<real_type>& opts;
> 
>   // only exposing the constructor and some methods
>   public:
>   {{ model_name }}(Args& args):
>     t(args.t),
>     pars(args.pars),
>     state_curr(args.state_curr),
>     state_next(args.state_next),
>     intermediate(args.intermediate),
>     opts(args.opts)
>   {};
> 
>   void run() { std::cout << "Running new model\n" };
> };
> ```
> 
> i do feel like this is probably one of the most confusing part of the step because theres concepts in there and template specialisation and stuff like that

Yes this is a good shout, we could even add a helper function for this. Instead of put it in the docs. That would take some user input on the command line and then write that file out. Could be quite nice! What do you think? A bit like one of the usethis functions.

I've added the template into the docs for now, but will make a ticket for helper function

M-Kusumgar (approved):
awesome! looks good to me, yhhh i think adding the helper function is the way to go!

PR #328, opened 2026-03-27
Add goals model

r-ash:
You can use this with the pickle file I sent over email. 

This is just a temporary measure to unblock you developing goals whilst we wait to wire up the data properly. What I am expecting is that eventually we will have an hdf5 file we can create in the same way as the other test data here, but from Python instead. But for now this should do.

The pickle file contains the output of running the following:

```
def modvars_to_numpy(tag, value):
    # Taken from FilterUtils.py
    if type(value) is list:
        try:
            if len(value) > 0 and ((type(value[0]) is dict) or (type(value[0]) is str) or (type(value[0]) is bool)):
                value = np.array(value, order='C')
            else:
                value = np.array(value, order='C', dtype=np.dtype(np.float64))

        except:
            print(f'Failed to convert list to numpy array {tag}')

    return value


def test_run_aim_model():
    pjnz = "C:\\Users\\Test\\Downloads\\Zimbabwe.PJNZ"
    with open(pjnz, "rb") as fh:
        buf = BytesIO(fh.read())
    projection, params, epp_files, shiny90 = GB_ImportProjection(buf)

    numpy_modvars = {tag: modvars_to_numpy(tag, value) for tag, value in projection.items()}

    with open("zimbabwe_modvars.pickle", "wb") as handle:
        pickle.dump(numpy_modvars, handle, protocol=pickle.HIGHEST_PROTOCOL)
```

So the pickle contains the modvars read from the Zimbabwe test PJNZ that John shared with us. I've asked him whether we can push any of this onto GitHub to make testing easier, but not heard back from him yet, so best to keep this private for now.

With this see the scrip `goals/run_goals.py` you can run this with uv by going to the goals directory and running `uv run run_goals.py`. You can also setup vscode to run this, which might be easier for debugging. Give me a shout if you want a hand getting this set up.

In that script before the call to `run_goals` you have 2 data structures which you'll need
1. `modvars` - the raw modvars read from the PJNZ (just converted into numpy arrays)
2. `lf_data` - the result of translating the data from modvars into leapfrog data structures. As you add more inputs to goals, we'll need to add new items into this dict. We should do this by expanding the data mapping code. I've copied this over for now into `src/leapfrog_mapping/leapfrog_data_mapping.py` but ultimately we will call this code from SpectrumEngine and remove the duplicated code here.

Let me know how you get on with this or if you have any questions.

Note that to use this you might need to update the path to your local zimbabwe modvars pickle file.


leapfrog-core/include/models/goals_simulation.hpp
```diff
@@ -54,28 +341,3289 @@ struct GoalsSimulation<Config> {
...
+      n_hv.mult_art(hd) = p_hv.epi_inf_mult_art(t) * p_hv.epi_infectiousness(INF_SYMPT_NO_ART);
+  }
+
+  for (int s = S_MALE; s <= S_FEMALE; ++s)
+      for (int rg = RG_LRH; rg <= RG_TOTAL1; ++rg)
+          for (int m = PREP_OralDaily; m <= PREP_PEP ; ++m)
+         {
+
+             i_hv.prep_effect(rg,s) +=  p_hv.prep_effectiveness(PREP_EFFECT, m)
+                                      * p_hv.prep_effectiveness(PREP_ADH, m)
```
r-ash: I think these `prep_effictiveness` can read from beyond the extends. This is defined as 

```
   "prep_effectiveness": {
      "num_type": "real_type",
      "dims": ["SS::nPREP+1",  "SS::RN_DurationMonths"]
    },
```

where `nPREP` is 9 and `RN_DurationMonths` is 4. But in the loop we are looping from `m = PREP_OralDaily; m <= PREP_PEP` from 0 to 9. So I think m can be greater than the number of `RN_DurationMonths` we have

r-ash:
Closing this as we've tidied the changes here and put them into commit https://github.com/hivtools/leapfrog/pull/334 and https://github.com/hivtools/leapfrog/pull/342

Issue #335, opened 2026-06-05
Paediatric model diverging when files use mid-year population projections


mwalte10:
Jeff raised that the Malawi file was diverging between the population level and paediatric stratified HIV/ART populations. Noted here: https://teams.microsoft.com/l/message/19:67778a9a51cf4510adce8cc6c7a5a1b7@thread.v2/1780584509180?context=%7B%22contextType%22%3A%22chat%22%7D

jeffeaton:
Thanks for isolating the issue on that.

That also raises another question / concern that I have about why that file is assigning as "mid-year" projection. It is a 2025 file, which should be calendar year. I am suspecting something in the PJNZ parser. I will look into that now.

But even with that, I think the simulation model should remain aligned for the mid-year projection.

-----
_Edit:_ This issue is fixed by #337. Example code for this error will need to manually specify version now:

```r
pkgbuild::clean_dll("leapfrogr/")
pkgbuild::compile_dll("leapfrogr/", debug = FALSE)
devtools::load_all("leapfrogr/")

pjnz <- "../Malawi_280525_FINAL.pjnz"
parameters <- process_pjnz(pjnz, extract_child_params = TRUE)
parameters$projection_period <- "midyear"

lsimP <- run_model(parameters, "ChildModel", 1970:2030)

p_hivpop_0to4 <- colSums(lsimP$p_hivpop[1:5,,],,2)
h_hivpop_0to4 <- colSums(lsimP$hc1_hivpop,,4) + colSums(lsimP$hc1_artpop,,4)

## Should be equal
round(cbind(p_hivpop_0to4,
            h_hivpop_0to4,
            p_hivpop_0to4 - h_hivpop_0to4), 4)


p_hivpop_5to14 <- colSums(lsimP$p_hivpop[6:15,,],,2)
h_hivpop_5to14 <- colSums(lsimP$hc2_hivpop,,4) + colSums(lsimP$hc2_artpop,,4)

## Should be equal
round(cbind(p_hivpop_5to14,
            h_hivpop_5to14,
            p_hivpop_5to14 - h_hivpop_5to14), 4)
        
```

Output:
```r
> devtools::load_all("leapfrogr/")
ℹ Loading leapfrog
> pjnz <- "../Malawi_280525_FINAL.pjnz"
> parameters <- process_pjnz(pjnz, extract_child_params = TRUE)
parameters$projection_period <- "midyear"
[...]

PR #338, opened 2026-06-05
Mkw issue 335

mwalte10:
Closes issue #335. 

Three changes implemented:

  1. Issue raised by Jeff: Fix child HIV population migration for midyear projections (leapfrog.hpp) run_hc_hivpop_end_year_migration() was inside a PROJPERIOD_CALENDAR-only guard. For midyear projections, net migration was applied to p_hivpop for child ages (in run_hiv_and_art_stratified_deaths_and_migration) but never to hc1_hivpop/hc2_hivpop, causing  the two population representations to diverge. The fix moves the child migration call outside the calendar-year guard so it runs unconditionally when the child model is active. This wasn't an issue for the test file as it didn't use mid year projections. 
  2. Fix accumulating deaths_migrate variable (hiv_demographic_projection.hpp) In run_hc_hiv_and_art_stratified_deaths_and_migration(), deaths_migrate was declared once before the age and sex loops and accumulated across every iteration. Each age/sex combination was therefore adding to a running total from prior iterations rather than computing its own value. The fix moves the declaration inside the inner loop so it resets to  zero per age/sex/year.
  3. Extend nosocomial infections to all three child age groups and rename parameter (child_model_simulation.hpp, config, R interface) nosocomial_infections() only populated hc1 (ages 0–4); the 5–9 and 10–14 rows of the input were silently ignored. The fix extends the function to apply infections to hc2 age groups as   well, using the correct hc2_hivpop index and += (previously =). Alongside this, the parameter is renamed from the ambiguous hc_nosocomial/nosocomial_infections_by_age to hc_nosocomial_infections_by_age throughout the codebase (config schema, generated  C++/R/Python adapters, R reader, and tests), and its shape changes from a 1-D vector to a 3×proj_steps matrix — one row per coarse child age group.

*Note that merge conflicts were resolved by me, and then investigation into Jeff's problem noted in #335 was mostly done by Claude.*


leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -656,17 +656,28 @@ struct ChildModelSimulation<Config> {
...
+    for (int s = 0; s < NS; ++s) {
+      for (int a = hc2_agestart; a < hcAG_end; ++a) {
+        int ag = (a - hc2_agestart) / 5 + 1;
+        if (p_hc.hc_nosocomial_infections_by_age(ag, t) > 0) {
+          auto infections = p_hc.hc_nosocomial_infections_by_age(ag, t) / (5.0 * NS);
+          n_ha.p_infections(a, s) += infections;
+          n_hc.hc2_hivpop(0, 0, a - hc2_agestart, s) += infections;
+        }
+      }
+    }
```
mwalte10, line 680: Kind of was using this PR to test claude's ability in this context, I'll go through and fix this to make it more aligned with the other coarse age group structures.

PR #339, opened 2026-06-09
Mkw issue 335 new

mwalte10:
Closes issue #335. 

Three changes implemented:

  1. Issue raised by Jeff: Fix child HIV population migration for midyear projections (leapfrog.hpp) run_hc_hivpop_end_year_migration() was inside a PROJPERIOD_CALENDAR-only guard. For midyear projections, net migration was applied to p_hivpop for child ages (in run_hiv_and_art_stratified_deaths_and_migration) but never to hc1_hivpop/hc2_hivpop, causing  the two population representations to diverge. The fix moves the child migration call outside the calendar-year guard so it runs unconditionally when the child model is active. This wasn't an issue for the test file as it didn't use mid year projections. 
  2. Fix accumulating deaths_migrate variable (hiv_demographic_projection.hpp) In run_hc_hiv_and_art_stratified_deaths_and_migration(), deaths_migrate was declared once before the age and sex loops and accumulated across every iteration. Each age/sex combination was therefore adding to a running total from prior iterations rather than computing its own value. The fix moves the declaration inside the inner loop so it resets to  zero per age/sex/year.
  3. Extend nosocomial infections to all three child age groups and rename parameter (child_model_simulation.hpp, config, R interface) nosocomial_infections() only populated hc1 (ages 0–4); the 5–9 and 10–14 rows of the input were silently ignored. The fix extends the function to apply infections to hc2 age groups as   well, using the correct hc2_hivpop index and += (previously =). Alongside this, the parameter is renamed from the ambiguous hc_nosocomial/nosocomial_infections_by_age to hc_nosocomial_infections_by_age throughout the codebase (config schema, generated  C++/R/Python adapters, R reader, and tests), and its shape changes from a 1-D vector to a 3×proj_steps matrix — one row per coarse child age group.

*Note that merge conflicts were resolved by me, and then investigation into Jeff's problem noted in #335 was mostly done by Claude.*

*Sorry for the _new in the branch name*


leapfrog-core/model_schemas/configs/HcConfig.json
```diff
@@ -47,9 +47,9 @@
     }
   },
   "pars": {
-      "hc_nosocomial": {
+      "hc_nosocomial_infections_by_age": {
         "num_type": "real_type",
-        "dims": ["opts.proj_steps"]
+        "dims": ["3", "opts.proj_steps"]
```
r-ash: Do we have a const for the 3? Or a new one we can add to the state space?

leapfrogr/R/process_pjnz_hc.R
```diff
@@ -413,8 +413,13 @@ process_pjnz_hc <- function(dat, pars, dim_vars, use_coarse_age_groups = FALSE,
   pars$breastfeeding_duration_art <- pars$infant_feeding_options[, "art", ] / 100
   pars$breastfeeding_duration_no_art <- pars$infant_feeding_options[, "no art", ] / 100
 
-  ##TODO: change this in leapfrog to be by multiple ages
-  pars$hc_nosocomial <- pars$nosocomial_infections_by_age[1, ]
+  ## Nosocomial infections by child age group (0-4, 5-9, 10-14); absent in older PJNZ files
+  if (!is.null(pars$nosocomial_infections_by_age)) {
+    pars$hc_nosocomial_infections_by_age <- pars$nosocomial_infections_by_age
+  }
```
r-ash: I don't think you need this chunk

leapfrogr/scripts/create_test_data.R
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env Rscript
+# Creates HDF5 parameter files used by leapfrogr tests.
+# Run from the root of the leapfrogr package:
+#   Rscript scripts/create_test_data.R
+
```
r-ash: This file is in the root of the project now, not "leapfrog/scripts/create_test_data.R" but "scripts/create_test_data.R"

Can you port any modifications over and remove this file

leapfrog.Rproj
```diff
@@ -0,0 +1,16 @@
+Version: 1.0
```
r-ash: Do you want to keep this? I'd rather remove it if it was accidental push, but if you want it so you can keep using RStudio when opening from the root that is fine. Have you tried switching to positron yet?

goals/tests/test_goals.py
```diff
@@ -21,6 +22,9 @@ def test_goals_model(test_data):
     modvars = modvars_to_numpy(modvars)
     ss = get_goals_ss()
     lf_data = modvars_to_leapfrog(modvars, ss, "Goals")
+    n_years = modvars[PJN_FinalYearTag] - modvars[PJN_FirstYearTag] + 1
+    if "hc_nosocomial_infections_by_age" not in lf_data:
+        lf_data["hc_nosocomial_infections_by_age"] = np.zeros((3, n_years))
```
r-ash, line 27: I'll update this one in Spectrum code

r-ash (changes requested):
Thanks Maggie, this is looking good, a few suggestions.

PR #348, opened 2026-06-18
Read RN_Type in for vaccine parameters (we need for goals-scenario)

r-ash:
This PR will:
1. Take RN_Type for vaccine parameters and pass it into model

We need to update the model to use this (somewhere?) and then expose this from goals-scenario. See the ticket on the board number 44


r-ash:
[Implement vaccine_action_type for vaccine parameters](https://trello.com/c/Qqzniap4/44-implement-vaccineactiontype-for-vaccine-parameters)

PR #349, opened 2026-06-19
Fix some of the code quality comments

r-ash:
From card48 this fixes:
1. `pop_reached` in `RN_CURE_NEO` case is immediately overwritten
2. `totalSexActs` in `calc_new_infections` is shadowed by another delcaration at line 2229
3. double value in `calc_resource_needs` only appears in commented out code, never used.
4. double literals should be replaced with `real_type`


r-ash:
[Code quality improvements](https://trello.com/c/HFr727ym/48-code-quality-improvements)

PR #350, opened 2026-06-19
Age 10-14 and sexual debut population implementation

jeffeaton:
This PR implements two main changes:

* Extends age range for adult HIV population from age 10 to 105 for GBD study and shifts the transition age from child to adult HIV model from age 15 to age 10.
* Add 'virgin' (not sexually active) population arrays for age 10-29 specifying a proportion of the population who are not exposed to HIV acquisition

I think we should aim to merge this to avoid code divergences.

Next steps for this feature implementation:

* Update fertility and MTCT model to only consider births among sexually active population 
* Update EPP-ASM transmission rate model to remove the virgin population from contributing to transmission (partially implemented, needs completion)
* Consider how to implement coarse age stratification version of the model for EPP fitting
  * Currently virgin population implementation assumes


leapfrog-core/include/models/hiv_demographic_projection.hpp
```diff
@@ -165,6 +165,115 @@ struct HivDemographicProjection<Config> {
...
+      n_vg.p_totpop_virgin(va, s) = c_dp.p_totpop(a - 1, s) * p_dp.survival_probability(a, s, t);
+      n_vg.p_hivpop_virgin(va, s) = c_ha.p_hivpop(a - 1, s) * p_dp.survival_probability(a, s, t);
+
+      // Entrants into virgin HIV stratified population
+      // Note: code relies on virgin HIV population being single-age stratified; needs
+      // to be generalised if implementing coarse HIV population.
+      for (int hm = 0; hm < hDS; ++hm) {
+        n_vg.h_hivpop_virgin(hm, va, s) += i_hc.age15_hivpop(hm, s) * p_dp.survival_probability(a, s, t);
+
+        if (t > p_hc.hc_art_start) {
```
r-ash, line 232: In some places we're using `t > p_hc.hc_art_start`. In the child model and in `run_hc_hivpop_end_year_migration` we're using `t >= p_hc.hc_art_start`. Just wanted to double check it is correct we have different conditionals? Or should the child model also be `t>`?

leapfrog-core/include/models/hiv_demographic_projection.hpp
```diff
@@ -165,6 +165,115 @@ struct HivDemographicProjection<Config> {
...
+          if (t > opts.ts_art_start) {
+            for (int hu = 0; hu < hTS; ++hu) {
+              n_vg.h_artpop_virgin(hu, hm, va, s) *= 1.0 + i_dp.migration_rate(a, s);
+            }
+          }
+        }
+      }
+
+      // Becoming sexually active
+      for (int va = 0; va < SS::vAG; ++va) {
```
r-ash, line 260: This shadows the `va` declared on line 221. Just double checking that is fine?

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -133,6 +152,17 @@ struct AdultHivModelSimulation<Config> {
       for (int a = p_ha.pIDX_INCIDPOP; a < p_ha.pIDX_INCIDPOP + p_ha.pAG_INCIDPOP; ++a) {
         i_ha.hiv_neg_aggregate(s) += c_dp.p_totpop(a, s) - c_ha.p_hivpop(a, s);
       }
+
+      if constexpr(ModelVariant::run_virgin) {
```
r-ash, line 156: The switches are starting to get many. From goals and new model variants here. 

At some point we might need to think about this, I wonder if there are ways we can make it a bit easier to see what bits of functionality different models will turn on/off.

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -296,21 +359,31 @@ struct AdultHivModelSimulation<Config> {
...
+      }
+
+      if constexpr(ModelVariant::run_virgin) {
+
+        // Remove virgin HIV negative from at risk population
+        const auto& n_vg = state_next.vg;
+
+        for (int va = 0, a = SS::p_idx_virginpop_first; va < SS::vAG; ++va, ++a) {
+          i_ha.hiv_negative_pop(a, s) -= n_vg.p_totpop_virgin(va, s) - n_vg.p_hivpop_virgin(va, s);
+        }
```
r-ash, line 377: We could put this into a function as we're doing the same thing here and in `calc_new_infections_incidmod_transmrate`.

leapfrog-core/model_schemas/configs/VgConfig.json
```diff
@@ -0,0 +1,50 @@
...
+    "overrides": [
+      {
+        "applies_if": "age10_adult",
+        "vars": {	      
+          "vAG": 20,
+          "p_idx_virginpop_first": 10
+        }
+      }	
+    ]
+  },
```
r-ash, line 20: Is there ever going to be a model with `run_virgin` true and `age10_adult` false?

r-ash (approved):
Looks good, I got made a couple of fixes:

1. Removed unpacking of `SS::vAG` in AdultHivModelSimulation. This wasn't compiling on windows machines. Some impl difference in how MSVC must work compared to gcc. I have removed the unpacking and we just use it name qualified. 
2. One of the goals functions was still using `hiv_negative_pop` with only age-stratified, I updated to use new age and sex stratified.
3. Some formatting updates

leapfrog-core/include/models/hiv_demographic_projection.hpp
jeffeaton: It is correct that it can be different. In general, it should depend on when it is called within the annual time relative to the ART initiation step. If it is before ART initiation step it should `>`, if it is after the ART initiation step (as the case for end year migration), it should be `>=`.

leapfrog-core/include/models/hiv_demographic_projection.hpp
jeffeaton: I'm not familiar with the terminology 'shadows', but from context I think you mean that variable `va` is already in scope here from above, and then redeclared here. [Let me know if not interpreted correctly.]

This is fine in terms of the code / algorithm doing the 'right thing' here, but probably poor form (?). The `int va = 0` on line 221 is essentially declared as a constant for clarity about what the code is doing there (vs. putting a 0 magic number in lines 222-257).  I will rename that `const int va0 = 0` to avoid clash.

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
jeffeaton: Agree

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
jeffeaton: Agree -- I've created an issue for that (#352)

leapfrog-core/model_schemas/configs/VgConfig.json
jeffeaton: I think that is TBC. I would like to leave that option open for the moment.

leapfrog-core/include/models/hiv_demographic_projection.hpp
r-ash: Yep, that is correct interpretation of "shadows". Good to remove them, it is an occasional source of bugs and is often a thing that linters will warn you about.

PR #354, opened 2026-06-25
Split hcAG_coarse into total and disaggregated data structures

### LINKED ISSUE #351: Review joined total and disaggregated inputs in child model
For eaxmple hcAG_coarse is stratified by total, 0-4, 5-9, 10-14. Let's instead input aggregated, total and an array of bools for which value to use. It makes the code a bit simpler to read. There are examples of this style in the adult model. Review other inputs and outputs too.

r-ash:
Closes #351 

This will make it more consistent with inputs in the adult model.

Ignore the goals tests for now, fixing those will require an update to Spectrum reading code to supply new split data, I'll get that updated once this is in.


leapfrog-core/model_schemas/configs/HcConfig.json
```diff
@@ -49,11 +48,11 @@
     "overrides": [
       {
         "applies_if": "age10_adult",
-        "vars": {	      
+        "vars": {
           "hcAG_end": 10,
           "hc2AG": 5,
-          "hcAG_coarse": 3,
-          "hc_age_coarse": [1, 1, 1, 1, 1, 2, 2, 2, 2, 2],
+          "hcAG_c": 2,
```
r-ash, line 54: @jeffeaton making this change to drop hcAG_coarse and use hcAG_c instead suggests me that this should be 2 here? But that was not in your original changes.

This is only used for `hc_nosocomial_infections` which is indexed using `hc_age_coarse` so it was fine. Updating this should just mean that drops an unused row from `hc_nosocomial_infections` when `age10_adult` is true.

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1230,12 +1226,9 @@ struct ChildModelSimulation<Config> {
       } // end a
     } // end NS
 
-    for (int ag = 1; ag < hcAG_coarse; ++ag) {
-      i_hc.on_art(0) += i_hc.on_art(ag);
-      i_hc.unmet_need(0) += i_hc.unmet_need(ag);
```
mwalte10, line 1242: Are these not needed anywhere else then? I kind of think that they might be needed in the instance of if the ART input for children switches from age aggregated to age dis aggregated or vice versa... worried that might be an edge case not captured in the current saved test data.

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1244,39 +1237,44 @@ struct ChildModelSimulation<Config> {
...
+                i_hc.art_last_year(hc_age_coarse[a]) += c_hc.hc2_artpop(dur, hd, a - hc2_agestart, s);
               }
             }
           }
         }
       }
 
-      i_hc.total_art_last_year(0) = 0.0;
-      for (int hca = 1; hca < hcAG_coarse; ++hca) {
-	      i_hc.total_art_last_year(0) += i_hc.total_art_last_year(hca);
```
mwalte10, line 1274: Same flag here, worried it might be needed if switching between ART input type.

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1244,39 +1237,44 @@ struct ChildModelSimulation<Config> {
...
+      real_type last_year_on_art_total = 0.0;
+      for (int hca = 0; hca < hcAG_c; ++hca) {
```
mwalte10: hcag rather than hca? As age is tracked with ag rather than a?

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1244,39 +1237,44 @@ struct ChildModelSimulation<Config> {
...
+        last_year_on_art_total += i_hc.art_last_year(hca);
+      }
```
mwalte10: I recognise the original names aren't clear, however I think if we are breaking apart the all children (index 0), then that intermediate name should reflect that the "total" is referring to all ages. art_last_year_hcag?

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1295,18 +1293,20 @@ struct ChildModelSimulation<Config> {
     } else {
       if (p_hc.hc_art_isperc(t - 1)) {
         // ART entered as percent last year so convert to number
-        i_hc.total_art_last_year(0) = p_hc.hc_art_val(0, t - 1) * i_hc.total_need(0);
+        real_type total_need_sum = 0.0;
+        for (int hca = 0; hca < hcAG_c; ++hca) {
+          total_need_sum += i_hc.total_need(hca);
+        }
```
mwalte10: Ah ok here is where those things above are coming in, same comment of would be nice if the the name reflected that it is all ages...

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1295,18 +1293,20 @@ struct ChildModelSimulation<Config> {
...
+        i_hc.art_last_year_total = p_hc.hc_art_val_total(t - 1) * total_need_sum;
```
mwalte10: So  p_hc.hc_art_val_total is a parameter that sums across p_hc.hc_art_val? Is there any check that makes sure those are consistent if they happen to be entered in the same file? (I actually don't think the Spectrum software will allow you to do that, but just want to be sure)

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1295,18 +1293,20 @@ struct ChildModelSimulation<Config> {
...
       } else if (p_hc.hc_art_is_age_spec(t - 1)) {
-        // ART entered as number last year but this year isn't then aggregate
-        // ages
-
-	      i_hc.total_art_last_year(0) = 0.0;
-	      for (int hca = 1; hca < hcAG_coarse; ++hca) {
-	        i_hc.total_art_last_year(0) += p_hc.hc_art_val(hca, t - 1);
-	      }
+        // ART entered as number last year but this year isn't then aggregate ages
```
mwalte10: My original comments grammar was tough, can you change to: ART entered as number in t-1 and dis-aggregated by age.

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1295,18 +1293,20 @@ struct ChildModelSimulation<Config> {
...
+        i_hc.art_last_year_total = 0.0;
+        for (int hca = 0; hca < hcAG_c; ++hca) {
+          i_hc.art_last_year_total += p_hc.hc_art_val(hca, t - 1);
+        }
       } else {
         // Last year was age aggregated and a number so use previous value
```
mwalte10: Change comment to: ART entered as number in t-1 and aggregated across ages.

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1335,13 +1346,25 @@ struct ChildModelSimulation<Config> {
     art_this_year();
 
     i_hc.retained = 1 - p_hc.hc_art_ltfu(t);
-    for (int ag = 0; ag < hcAG_coarse; ++ag) {
-      auto average_art_by_year = (i_hc.total_art_last_year(ag) + i_hc.total_art_this_year(ag)) /
-                                2.0;
-      n_hc.hc_art_init(ag) = std::max(i_hc.hc_art_deaths(ag) + average_art_by_year - i_hc.on_art(ag) * i_hc.retained, 0.0);
-      n_hc.hc_art_init(ag) = std::min(n_hc.hc_art_init(ag),
-                                      i_hc.unmet_need(ag) + i_hc.on_art(ag) * p_hc.hc_art_ltfu(t));
-    } // end ag
```
mwalte10, line 1351: Issue here was that it wasn't consistent with the adult model? To me this reads a lot easier but not a hill I'll die on

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1502,34 +1525,38 @@ struct ChildModelSimulation<Config> {
         } // end a
       } // end  NS
     } else {
+      real_type hc_initByAge_total = 0.0;
       for (int s = 0; s < NS; ++s) {
         for (int a = 0; a < hcAG_end; ++a) {
           for (int cat = 0; cat < hcTT; ++cat) {
             for (int hd = 0; hd < hc1DS; ++hd) {
-              i_hc.hc_initByAge(0) += n_hc.hc_art_need_init(hd, cat, a, s) * p_hc.hc_art_init_dist(a, t);
+              hc_initByAge_total += n_hc.hc_art_need_init(hd, cat, a, s) * p_hc.hc_art_init_dist(a, t);
```
mwalte10, line 1541: same flag on _total

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1502,34 +1525,38 @@ struct ChildModelSimulation<Config> {
...
       }
 
+      auto hc_art_val_sum = p_hc.hc_art_val_total(t) + p_hc.hc_art_val_total(t - 1);
+
       for (int s = 0; s < NS; ++s) {
         for (int a = 0; a < hcAG_end; ++a) {
           for (int cat = 0; cat < hcTT; ++cat) {
             for (int hd = 0; hd < hc1DS; ++hd) {
-              auto hc_art_val_sum = p_hc.hc_art_val(0, t) + p_hc.hc_art_val(0, t - 1);
+              real_type hc_art_scalar_total;
```
mwalte10, line 1560: _total note

leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -1502,34 +1525,38 @@ struct ChildModelSimulation<Config> {
...
             } // end hc1DS
           } // end hcTT
         } // end a
       } // end  NS
 
-      if (i_hc.hc_initByAge(0) == 0.0) {
-        i_hc.hc_adj(0) = 1.0 ;
+      real_type hc_adj_total;
+      if (hc_initByAge_total == 0.0) {
+        hc_adj_total = 1.0;
```
mwalte10, line 1549: _total note

leapfrogr/R/process_pjnz_hc.R
```diff
@@ -403,7 +403,7 @@ process_pjnz_hc <- function(dat, pars, dim_vars, dp_params, use_coarse_age_group
   if (!is.null(pars$hc_nosocomial_infections_by_age)) {
     hc_nosocomial_infections_by_age <- pars$hc_nosocomial_infections_by_age
   } else {
-    hc_nosocomial_infections_by_age <- matrix(0, nrow = 3, ncol = length(proj_years))
```
mwalte10, line 406: @r-ash  can you confirm there is no way to put in a 0-14 value for nosocomial infections? I think some older iteration of spectrum files had that as an option

mwalte10 (changes requested):
I have a few suggestions of edge cases to check as i remember when coding this I would get the code to work in one pattern of ART entries and then it would instantly break. 

Would be good to test that this survives if you: 
go from 0-14 number entry to age disaggregate number entry
age disaggregate numer entry to 0-14% entry. 
Those two patterns are what the vast majority of spectrum files will have.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: `hcag` sounds good, I'll update that.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: Let's chat about briefly in meeting today and agree on a name.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: So we do need these, but there are new separate intermediate variables for it, or we just calculate them as a local variable if they are only needed once.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: Actually we use `ag` elsewhere in this file, I will switch to that.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: No it was, this should be doing exactly the same thing for the case with disaggregated ages. For total, previously it could just use the same logic with index 0. But now the names of the parameters and intermediate variables are a bit different, so we need to handle the "total" separately.

One advantage is we can now have the extra conditional on `hc_art_is_age_spec` (I think?) because in the case that it is not age_spec this would have just been setting `hc_art_init` to 0 for index 1, 2, and 3. But it will be 0 anyway. So I think we can use the switch to save us a bit of work.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: Not sure, and I'm not sure where they are entered either. Let's review in meeting today

leapfrogr/R/process_pjnz_hc.R
r-ash: According to the PJNZ reading code the first tag we have there just reads in disaggregated https://github.com/hivtools/pjnz/blob/6478f6c43870f37dd1c79cfc7d5f9afc8adc498d/R/pjnz_metadata.R#L873 It would be good to double check with Rob too.

However I think you have made me notice a bug, this input code is expecting that parameter to be called `hc_nosocomial_infections_by_age` but I think PJNZ is reading it in as `nosocomial_infections_by_age`. I will fix that

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: They can't be from Spectrum, but we should consider users running this directly. They can set a flag for whether to use disaggregated or aggregated value. I will check that we favour the disaggregated when both are entered.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: Let's check the edge cases, and make sure that we don't need to compute the total always.

leapfrogr/R/process_pjnz_hc.R
r-ash: We did used to take aggregated age nosocomial infections. Rob G is going to review how they were used and get back to me.

leapfrog-core/model_schemas/configs/HcConfig.json
```diff
@@ -235,43 +238,49 @@
...
-      "dims": ["SS::hcAG_coarse"]
+      "dims": ["SS::hcAG_c"]
     },
     "on_art": {
       "num_type": "real_type",
-      "dims": ["SS::hcAG_coarse"]
+      "dims": ["SS::hcAG_c"]
     },
-    "total_art_last_year": {
+    "art_last_year": {
```
r-ash: `art_previous`

leapfrog-core/model_schemas/configs/HcConfig.json
```diff
@@ -235,43 +238,49 @@
...
+    "art_last_year_total": {
+      "num_type": "real_type"
     },
-    "total_art_this_year": {
+    "art_this_year": {
       "num_type": "real_type",
-      "dims": ["SS::hcAG_coarse"]
+      "dims": ["SS::hcAG_c"]
+    },
+    "art_this_year_total": {
```
r-ash: `art_current_total`

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: Discussed in meeting, Jeff wants to keep `total`,

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: Discussed in meeting, Jeff wants to keep `total`.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: Discussed in meeting, Jeff wants to keep `total`.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: Replaced this with `total_need_all_children`

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: Replaced this with `on_art_previous_total` which matches other names from Jeff.

leapfrogr/R/process_pjnz_hc.R
r-ash: So Rob G confirmed in an email on Friday 26th

> 1. Before Spectrum 6.09, we used `<NosocomialInfections MV>`. This specified nosocomial infections by year with no further disaggregation. Nosocomial infections were added to the compartment of age 0 perinatal infections.
> 2. Spectrum 6.09 introduced a new tag for age-disaggregated inputs, `<NosocomialInfectionsByAge MV>`. This tag is still current. This version disaggregated nosocomial infections from five-year age groups to single ages proportional to the age distribution among existing CLHIV.
> 3. Spectrum 6.18 changed to the current disaggregation scheme, which distributes nosocomial infections by five-year age group uniformly by single ages.
> 
> Spectrum versions 6.09 and 6.18 were released in November 2021 and March 2022 respectively, so both part of the 2022 estimates round.

I will follow up with him what we do about 2 vs 3: In leapfrog we are distributing them uniformly by single ages. Do we need to support case 2?

For case 1, I can update the `process_pjnz_hc` so that if `NosocomialInfectionsByAge MV` tag is missing from the PJNZ then we pass in `NoscomialInfections MV` into the model and only apply for age 0. 

I think there is a general question here about if we support previous versions of Spectrum engine when Spectrum itself does not? What is our general policy? We try to keep the PJNZ reading compatible with old versions, but do we need to keep old versions of the model running. That sounds very difficult.

r-ash: So we decided to

1. Change nosocomial infections to be input by single age
2. Support the pre 6.09 version which put all nosocomial infections into the age 0 compartment
3. Not support between 6.09 and 6.18, it should be unlikely we get PJNZ files from this era.

I have made these changes.

r-ash:
> I have a few suggestions of edge cases to check as i remember when coding this I would get the code to work in one pattern of ART entries and then it would instantly break.
> 
> Would be good to test that this survives if you: go from 0-14 number entry to age disaggregate number entry age disaggregate numer entry to 0-14% entry. Those two patterns are what the vast majority of spectrum files will have.

Thanks Maggie, I tested this by creating a file with the following child ART input:
1. number -> percent (year 2022)
2. percent -> number (year 2017 and 2021)
3. number -> number age disaggregated (2018)
4. % -> number age disaggregated (2030)
5. number age disaggregated -> number (2031)
6. number age disaggregated -> % (2019)

And I could see this was breaking when t-1 was a non-age specific year as we we're only accumulating from the age disaggregated and not the total.

I've now fixed this and in comparing between this branch and current main the numbers agree up to floating point precision. So I think safe that this PR does not alter the calculations.

mwalte10 (approved):
Looks good!

PR #356, opened 2026-07-09
HaConfig: added art_initiation_rate

CarelPretorius:
Adding functionality to calc ART initiation directly from a input rate

HaConfig: art_initiation_rate
adult_sim: added code to use art initiation rate to calc new art initiation
goals_sim: : added code to use art initiation rate to calc new art initiation
HvConfig: renamed an input parameter


leapfrog-core/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -521,7 +521,7 @@ struct AdultHivModelSimulation<Config> {
                real_type temp_art_adult_dropout = 0.0;
                if constexpr (ModelVariant::run_goals) {
                    if(t > pars.hv.goals_base_year_idx){ 
-                    temp_art_adult_dropout = -std::log(1.0 - pars.hv.art_interupt_rate(t) *
+                    temp_art_adult_dropout = -std::log(1.0 - pars.hv.art_interrupt_rate(t) *
```
CarelPretorius, line 524: changed the name tpo: art_interrupt_rate

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -548,99 +548,122 @@ struct AdultHivModelSimulation<Config> {
...
+        for (int hm = i_ha.everARTelig_idx; hm < hDS; ++hm) {
+            for (int hu = 0; hu < hTS; ++hu) {
+              OnART += c_ha.h_artpop(hu, hm, ha, s);
+            }
+        }
+      }
+      
+      i_ha.artinit_hts = std::max( opts.dt * pars.ha.art_initiation_rate(s,t) * (NewInf + PLHIV - OnART), 0.0);
+    }
+    else{
```
CarelPretorius: calculating ART initiation rate directly from input. The previous calc of new ART is in the second block of the iff statement.  if (pars.hv.art_cov_num_percent[0] == ART_INIT_RATE) . A SS variable can be created for  int ART_INIT_RATE = 5;

leapfrog-core/model_schemas/configs/HaConfig.json
```diff
@@ -218,7 +218,14 @@
     "pwid_sex_ratio": {
       "num_type": "real_type",
       "dims": ["opts.proj_steps"]
+    },
+
+    "art_initiation_rate": {
+      "num_type": "real_type",
+      "dims": ["SS::NS", "opts.proj_steps"]
```
CarelPretorius, line 227: added this input var

leapfrog-core/model_schemas/configs/HvConfig.json
```diff
@@ -332,7 +332,7 @@
       "num_type": "real_type"
     },
    
-    "art_interupt_rate": {
+    "art_interrupt_rate": {
```
CarelPretorius, line 329: corrected the var name

leapfrog-core/include/models/goals_simulation.hpp
```diff
@@ -2951,10 +2953,25 @@ struct GoalsSimulation<Config> {
...
+
+            real_type NewInf = c_hv.new_inf_vrs(v, rg, s); 
+            real_type PLHIV = 0.0;
+            real_type OnART = 0.0;
+            for (int hd = CD4_GT500; hd <= CD4_LT50; ++hd) {
+              PLHIV += c_hv.adults(v, rg, hd, s);
+              OnART += c_hv.adults(v, rg, hd + hOnArt, s);
+            }
+
+            start_art[v][rg][s] = std::max( opts.dt * pars.ha.art_initiation_rate(s,t) * (NewInf + PLHIV - OnART), 0.0);
```
CarelPretorius: calculating new ART directly from an input rate, instead of from a coverage

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -548,99 +548,122 @@ struct AdultHivModelSimulation<Config> {
...
   void run_h_art_initiation(int hiv_step) {
     const auto& p_ha = pars.ha;
+    const auto& c_ha = state_curr.ha;
     auto& n_ha = state_next.ha;
     auto& i_ha = intermediate.ha;
 
     for (int s = 0; s < NS; ++s) {
-      i_ha.Xart_15plus = 0.0;
+      
+      int ART_INIT_RATE = 5;
```
r-ash: You can make this `static constexpr`, this will ensure that ART_INIT_RATE is created at compile time and always refers to the same variable (no allocation within the loop)

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -548,99 +548,122 @@ struct AdultHivModelSimulation<Config> {
...
+      real_type OnART = 0.0;
+      for (int ha = 0; ha < hAG; ++ha) {
+        const int a = ha + SS::p_idx_hiv_first_adult;
+        NewInf +=  c_ha.p_infections(a, s);       
+        PLHIV += c_ha.p_hivpop(a, s);             
+        for (int hm = i_ha.everARTelig_idx; hm < hDS; ++hm) {
+            for (int hu = 0; hu < hTS; ++hu) {
+              OnART += c_ha.h_artpop(hu, hm, ha, s);
+            }
+        }
```
r-ash: Can we fix the indentation here please.

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
r-ash: Perhaps we should have a chat about this though, is there any reason we keep this as value 5? We have 3 cases?

1. Passed in art cov as number
2. Passed in art cov as %
3. Passed in art initiation rate

Can we mix between art initiation rate per year?

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -548,99 +548,122 @@ struct AdultHivModelSimulation<Config> {
...
     const auto& p_ha = pars.ha;
+    const auto& c_ha = state_curr.ha;
     auto& n_ha = state_next.ha;
     auto& i_ha = intermediate.ha;
 
     for (int s = 0; s < NS; ++s) {
-      i_ha.Xart_15plus = 0.0;
+      
+      int ART_INIT_RATE = 5; 
+      if (pars.hv.art_cov_num_percent[0] == ART_INIT_RATE) {
```
r-ash: We should change this name, it is misleading. In fact this won't work to refer to some data from the goals model here in the adult model sim.

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
```diff
@@ -919,9 +942,9 @@ struct AdultHivModelSimulation<Config> {
 
      for (int s = 0; s < NS; ++s) {
       cured_proportion =  intermediate.hv.cure_effect(s);
-      for (int ha = SS::p_idx_hiv_first_adult; ha < hAG; ++ha) {
+      for (int ha = 0; ha < hAG; ++ha) {
           
-          const int a = ha + p_idx_hiv_first_adult;
+          const int a = ha + SS::p_idx_hiv_first_adult;
           //adults, plhiv not on art
```
r-ash: Was this an intended change?

r-ash (changes requested):

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
r-ash: We pass this in with length 5 but then only ever use the 0th index of it? What is this input?

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
r-ash: Perhaps we could also use a real enum here

r-ash (changes requested):
This function is getting a bit big, I wonder if we try and refactor it to make it a bit easier to read, extract the if/else block into 2 helper functions `computeArtInitFromRate(..)` and `computeArtInitFromCoverage(..)` or similar?

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
CarelPretorius: yes

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
CarelPretorius: It has 5 potential values all in [0] since it comes from a radio button control group. This works for now but we can also change the input for to make it less confusing

CarelPretorius: happy to change the name pars.hv.art_cov_num_percent

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
CarelPretorius: ok

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
CarelPretorius: addressed in slack message: should work for now but a more general implementation can follow

leapfrog-core/include/models/goals_simulation.hpp
rlglaubius: Do we actually need to tally then deduct `OnART` here? It looks like `PLHIV` might already include just people not on ART.

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
rlglaubius: Is the conversion of the interruption rate from a percent to a per-capita rate via the log call needed here? Leapfrog might already handle that as a preprocessing step.

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
rlglaubius: Using `ART_INIT_RATE = 5` is an artifact from the Delphi implementation that exists because of the various ways adult ART uptake can be specified in Spectrum and the chronological order in which those were added to the software. It would be best to hide or abstract these artifacts from Leapfrog, but I defer to folks who know Leapfrog better what approach makes the most sense.

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
rlglaubius: I agree with Rob here, the `art_cov_num_percent` name didn't anticipate all the various ways we can specify ART coverage, so it seems best to make the name more general, e.g., `art_entry_option`.

rlglaubius (commented):
I've left a few comments and questions on some of the open discussions here.

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
r-ash: I've tidied this up and updated the input reader in SpectrumEngine

leapfrog-core/include/models/adult_hiv_model_simulation.hpp
r-ash: I've updated this, it is not `art_entry_option`.

r-ash:
[Support intervention ramp-up](https://trello.com/c/2L7fgtvC/53-support-intervention-ramp-up)

r-ash (approved):

PR #359, opened 2026-08-10
PMTCT need output

r-ash:
This makes PMTCT need an output. We want this to calculate prevalence among pregnant women for doing model comparisons with Spectrum.

### DIFF
--- a/leapfrog-core/include/models/child_model_simulation.hpp
+++ b/leapfrog-core/include/models/child_model_simulation.hpp
@@ -275,24 +275,24 @@ struct ChildModelSimulation<Config> {
       i_hc.sumARV += p_hc.PMTCT(hp, t);
     }
 
-    i_hc.need_PMTCT = std::max(i_hc.sumARV, n_ha.hiv_births);
+    n_hc.pmtct_need = std::max(i_hc.sumARV, n_ha.hiv_births);
 
     i_hc.on_PMTCT = i_hc.sumARV + p_hc.patients_reallocated(t);
-    i_hc.on_PMTCT = std::min(i_hc.on_PMTCT, i_hc.need_PMTCT);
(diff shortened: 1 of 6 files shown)


leapfrog-core/include/models/child_model_simulation.hpp
```diff
@@ -275,24 +275,24 @@ struct ChildModelSimulation<Config> {
       i_hc.sumARV += p_hc.PMTCT(hp, t);
     }
 
-    i_hc.need_PMTCT = std::max(i_hc.sumARV, n_ha.hiv_births);
+    n_hc.pmtct_need = std::max(i_hc.sumARV, n_ha.hiv_births);
 
     i_hc.on_PMTCT = i_hc.sumARV + p_hc.patients_reallocated(t);
-    i_hc.on_PMTCT = std::min(i_hc.on_PMTCT, i_hc.need_PMTCT);
+    i_hc.on_PMTCT = std::min(i_hc.on_PMTCT, n_hc.pmtct_need);
```
jeffeaton, line 281: Can this be not capitalised? Fine to leave that for another PR. [Seems like a great PR for Claude]

jeffeaton (approved):
This looks good to me.

We also want this output by age for other things we plan to do. That might be a bit more involved. Fine to merge this and leave that for later. But perhaps make a note (mental or otherwise) that might need to change any follow-on elements to sum over an age stratified array later.

leapfrog-core/include/models/child_model_simulation.hpp
r-ash: It can, the other outputs are consistently lower case e.g. `pmtct_coverage_at_delivery`, `mtct_by_source_hc_infections`, `hc1_art_aids_deaths` etc. I generally prefer to follow snake_case naming and keep it lower case for variable names. But could go the other way if you feel strongly, as long as we are consistent.

Issue #364, opened 2026-08-11
Remove unused Option A/B breastfeeding transmission calculations


jeffeaton:
The `adjust_option_A_B_bf_tr()` helper calculates `excessratio_bf`, `optA_bf_transmission_rate`, and `optB_bf_transmission_rate`, but none of these values are subsequently read.

Relevant code:
https://github.com/hivtools/leapfrog/blob/66b296cb6502fc8d02f65f1349ba8ba9bfb290a7/leapfrog-core/include/models/child_model_simulation.hpp#L397-L423

The function also repeats `calc_wlhiv_cd4_proportion()`, which has already run through `adjust_option_A_B_tr()`. Issue #274 notes that the alternative Option A/B transmission rates are not used.

Before deleting this code, confirm that these rates are not intended to feed the breastfeeding transmission calculation.

## Acceptance criteria

- Resolve the intended relationship with #274.
- If unused, remove `adjust_option_A_B_bf_tr()`, its call, and the three intermediate fields.
- If intended for use, wire the calculated rates into breastfeeding transmission and add coverage for that behavior.
- Confirm model-output equality across supported interfaces.

r-ash:
Related to #323 ?

PR #365, opened 2026-08-11
Localize scalar intermediates

### LINKED ISSUE #299: declaration of scalars as intermediates -- why?

jeffeaton:
_Created by Codex_

Closes #299

Follow-up issue #364 tracks the potentially dead Option A/B breastfeeding transmission calculations, which remain unchanged in this PR.

## Summary

- Replace function-local scalar intermediate members with narrowly scoped C++ locals.
- Explicitly initialize accumulators that previously relied on `Intermediate::reset()`.
- Remove unused and write-only scalar declarations unrelated to the restored breastfeeding calculation.
- Retain scalar intermediates whose values cross function boundaries.
- Reconcile the refactor with the latest `main`, retaining newly shared ART and Goals fields.

## Validation

- Code generation completed successfully.
- Generated schemas contain the expected scalar intermediate inventory.
- No removed intermediate member references remain.
- C++20 syntax check passes.
- Python extension compilation succeeds; the fixture-independent test passes.
- R package compilation and installation succeed.

## Test limitations

- Five Python model tests stop before model execution because the checked-in HDF5 fixtures lack parameters now required by `main`: `pwid_hivpos_nonaids_mortality` and `hc_nosocomial_infections_by_age`.
- Standalone CMake validation could not run because CMake is not installed locally.


r-ash (approved):
Beautiful

PR #376, opened 2026-08-14
Fix issue where FOI formula was setting transmission probabilty to 0

r-ash:
This was introduced after adding functional cure. Functional cure impact on infectiousness was originally added as a direct multipler. But changed to a (1.0 - func_cure_impact_inf) at some point. Its default value was correctpy updated in one place but not in init_vars_pre_hiv_loop. So it was setting transmission probability to 0.


r-ash:
This got pulled into #377

